PR #7, opened 2018-06-14
Added instructions for what would need to be done to start op-scim on startup via docker-compose

rickfillion:


docker-compose/README.md
```diff
@@ -29,4 +29,14 @@ Next, edit `docker-compose.yml`, replacing `{YOUR-DOMAIN-HERE}` with the domain
...
\ No newline at end of file
+Once you have tested the configuration, the bridge can be exited using ctrl/cmd-c, and restarted in daemon mode using `docker-compose up -d, or deployed for production use with Docker Swarm or Kubernetes. You can access logs using `docker-compose logs` at any point in the future.
+
+## Automatically starting the SCIM bridge upon startup
+
+In order to automatically start the SCIM bridge upon startup when using docker-compose you'll need to automatically start the Docker daemon, then start op-scim.
+
+### Systemd
+
+* Enable dockerd to run on startup: `systemctl enable dockerd`
```
ag-andrey05: i would re-order things in here, saying:
when systemd service (unit) file is configured reload systemd:
`systemctl daemon-reload`
and then enable the service with:
`systemctl enable op-scim`

docker-compose/op-scim.service
```diff
@@ -0,0 +1,17 @@
+#/etc/systemd/system/op-scim.service
```
ag-andrey05: I would change path to systemd unit file to `/lib/systemd/system/op-scim.service`

docker-compose/op-scim.service
```diff
@@ -0,0 +1,17 @@
+#/etc/systemd/system/op-scim.service 
+[Unit]
+Description=1Password SCIM bridge
+Requires=docker.service
+After=docker.service
+
+[Service]
+Type=oneshot
```
ag-andrey05: Type=simple

docker-compose/op-scim.service
```diff
@@ -0,0 +1,17 @@
...
+RemainAfterExit=yes
+WorkingDirectory=/path/to/directory/containing/docker-compose.yml
+ExecStart=/usr/local/bin/docker-compose up -d
```
ag-andrey05: can be used with `-f` to specify yml file

docker-compose/op-scim.service
rickfillion: this would depend on the distribution i imagine. on the CentOS VM that I'm playing with it's stored where I have it commented.

docker-compose/op-scim.service
rickfillion: thanks. since we name our example docker-compose.yml i think it's fine as is.

docker-compose/README.md
rickfillion: Not 100% sure I follow. do you mean tht the steps should be:

1. Install the service file (with edits)
2. reload systemd: `systemctl daemon-reload`
3. enable the op-scim service: `systemctl enable op-scim` 
?

ag-andrey05: yes, because whenever we make changes to service file or adding a new one systemd needs to re-read changes

rickfillion: thanks. will adjust.

PR #11, opened 2018-07-13
Fixed container command in op-scim-deployment.yaml

msweetser-isp:
When attempting to deploy op-scim, I received the following error:

```
    Last State:  Terminated
      Reason:    ContainerCannotRun
      Message:   oci runtime error: container_linux.go:247: starting container process caused "exec: \"./op-scim\": permission denied"
```

Changing the container command from ./op-scim to /op-scim/op-scim, as in the Docker compose file, fixed the issue.


agilecohix:
@msweetser-isp great catch, thank you for the PR!

PR #24, opened 2019-01-27
Nkechi/doc review

ogbuaguno:
1. Added a new script - `build.sh` - to run the commands required after the scam setup has been run.
1. Rewrote the docker swarm and docker compose README files to reflect new changes.


cohix:
I think with this script, we can just combine the two docker examples into one. Rather than `docker compose` and `docker swarm`, let's just have `docker` and have a paragraph at the beginning that explains the difference between compose and swarm (i.e. explain that compose should only be used for testing unless you know the implications, and that the docker host needs to be put in swarm mode to use swarm)

Otherwise this is really awesome!

build.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
cohix: This is great. Could you rename this to `deploy.sh` ?

ogbuaguno: got it

PR #73, opened 2019-07-04
Make sure user, group, and permissions are set

wayneworkman:
This fixes the installation on the latest available official Ubuntu 18.04 AMI, and maybe other systems too.


grellyd:
Thanks for the PR @wayneworkman. Sorry for the slow response. We don't monitor this repo closely for PRs. 

What led you to needing to move from a variable based port to hard coding the value? Was the value not properly propagated for some reason?

PR #94, opened 2020-04-29
Rewrite of the Docker deploy.sh script

alicethorne-ab:
This rewrite makes the script more robust and allows it to be executed from any directory, as well as avoiding any copy/move/delete commands, and removes the need to have separate *.sh files to generate secrets.

### DIFF
--- a/docker/deploy.sh
+++ b/docker/deploy.sh
@@ -16,7 +16,7 @@ run_docker_compose() {
     SESSION=$(cat $scimsession_file | base64 | tr -d "\n")
     echo "OP_SESSION=$SESSION" > $docker_path/$docker_type/scim.env
 
-    if ! docker-compose -f $docker_deploy_file up --build -d
+    if ! docker-compose -f $docker_file up --build -d
     then
         echo " "
         echo "Failed to run docker-compose, please investigate the error"
@@ -31,7 +31,7 @@ run_docker_compose() {
         echo " "
         echo "Press Ctrl+C to quit out of the log view."
         sleep 2
-        docker-compose -f $docker_deploy_file logs -f 2>/dev/null
+        docker-compose -f $docker_file logs -f 2>/dev/null
     else
         echo "Skipping logs..."
         echo "You can view the logs manually by running: docker-compose logs -f"
@@ -54,7 +54,7 @@ run_docker_swarm() {
         exit 1
     fi
 
-    if ! docker stack deploy -c $docker_deploy_file op-scim
+    if ! docker stack deploy -c $docker_file op-scim
     then
         echo " "
         echo "Failed to deploy to Docker Swarm, please investigate the error"
@@ -118,7 +118,7 @@ echo "Deployment type:" $docker_type
 echo "scimsession file path:" $scimsession_file
 echo "Domain name:" $domain_name
 
-while ! [[ "$proceed" =~ ^([yY][eE][sS]|[yY])$ || "$proceed" =~ ^([nN][oO][nN]) ]]; do
+while ! [[ "$proceed" =~ ^([yY][eE][sS]|[yY])$ ]]; do
     read -p "Does this look correct? [Y/n]: " proceed
     if [[ "$proceed" =~ ^([nN][oO][nN])$ ]]
     then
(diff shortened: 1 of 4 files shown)


docker/deploy.sh
```diff
@@ -1,56 +1,127 @@
...
+
+    if ! docker-compose -f $docker_deploy_file up --build -d
+    then
+        echo " "
+        echo "Failed to run docker-compose, please investigate the error"
+        sleep 1
+        exit 1
+    fi
+
+    while ! [[ "$view_logs" =~ ^([yY][eE][sS]|[yY])$ || "$view_logs" =~ ^([nN][oO][nN]) ]]; do
```
grellyd, line 67: Can you explain the presence of this conditional to me? Why is the loop needed given the `if` statements below?

docker/deploy.sh
```diff
@@ -1,56 +1,127 @@
...
+
+    if ! docker stack deploy -c $docker_deploy_file op-scim
+    then
+        echo " "
+        echo "Failed to deploy to Docker Swarm, please investigate the error"
+        sleep 1
+        exit 1
+    fi
+
+    while ! [[ "$view_logs" =~ ^([yY][eE][sS]|[yY])$ || "$view_logs" =~ ^([nN][oO][nN]) ]]; do
```
grellyd, line 106: Ditto as on Compose.

docker/deploy.sh
```diff
@@ -1,56 +1,127 @@
...
+    then
+        echo "Exiting..."
+        exit 1
+    fi
+done
+
+# place the domain name into the deployment file, in a backup
+docker_file_path=$docker_path/$docker_type
+docker_original_file=$docker_file_path/docker-compose.yml
+docker_deploy_file=$docker_file_path/docker-compose.yml.deploy
```
grellyd, line 45: Just to be clear, having the ending not strictly be a `.yml` and instead be a `.yml.deploy` does not interfere with the docker tools, correct?

alicethorne-ab: Works just fine if you're specifying the file by explicit name.

docker/deploy.sh
```diff
@@ -1,56 +1,150 @@
...
-    echo "Invalid docker manager. Please use either docker-compose or docker-swarm."
-    exit
 fi
+
+echo " "
+echo "Deployment of the 1Password SCIM Bridge is complete!"
+echo " "
+echo "If you have any issues deploying the SCIM Bridge, please either reach out to 1Password Business Support, or look through our helpful discussion forums: https://discussions.agilebits.com/categories/scim-bridge"
+echo " "
+echo "Back up of deployment docker-compose.yml file saved to" $docker_deploy_file
```
grellyd, line 150: If you are `sed s/{YOUR-DOMAIN-HERE}/$domain_name/g $docker_original_file > $docker_deploy_file` is the `$docker_deploy_file` really a backup? Do you mean the `#docker_original_file`?

docker/deploy.sh
```diff
@@ -1,56 +1,150 @@
...
+    echo "File '$scimsession_file' does not exist at that path, please try again." >&2
+done
+
+echo " "
+echo "Using the following parameters to deploy the SCIM Bridge"
+echo "Deployment type:" $docker_type
+echo "scimsession file path:" $scimsession_file
+echo "Domain name:" $domain_name
+
+while ! [[ "$proceed" =~ ^([yY][eE][sS]|[yY])$ || "$proceed" =~ ^([nN][oO][nN]) ]]; do
```
grellyd, line 121: As the loop exit clause only gets evaluated at the end of the loop, and `no` is checked for within the loop, isn't the `no` on the exit clause redundant?

docker/deploy.sh
alicethorne-ab: The idea here is that I'm point them towards the file that was actually used for their deployment, which is `$docker_deploy_file`. 

I'm realizing now that this might not be clear between the filenames the documentation use vs what's used in the script. I'll change that up.

docker/deploy.sh
alicethorne-ab: Fixed.

PR #97, opened 2020-05-08
Additional clarification of caveats section of PREPARING

alicethorne-ab:
Received an email asking for clarification of this section. Hopefully, this makes it more clear going forward.

### DIFF
--- a/PREPARATION.md
+++ b/PREPARATION.md
@@ -57,7 +57,7 @@ There are a few common issues that pop up when deploying the SCIM Bridge.
 * You should **never** need to log into this Provision Manager account manually. Please refrain from doing so.
 * Do not attempt to perform a provisioning sync until the setup has been completed.
 * Once set up, your Identity Provider becomes the _authoritative source_ of information for your 1Password accounts. With Provisioning enabled, the ability to change the _display name_ and _account status_ are not possible through the 1Password Web UI, and must be done through your Identity Provider. You can, however, continue to issue Account Recovery requests through the 1Password Web UI with Provisioning enabled.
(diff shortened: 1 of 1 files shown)


PREPARATION.md
```diff
@@ -56,7 +56,8 @@ There are a few common issues that pop up when deploying the SCIM Bridge.
...
 * You should **never** need to log into this Provision Manager account manually. Please refrain from doing so.
 * Do not attempt to perform a provisioning sync until the setup has been completed.
-* Once set up, your Identity Provider becomes the _authoritative source_ of information for your 1Password accounts. With Provisioning enabled, you _cannot edit user details manually_, and must do so through your Identity Provider.
+* Once set up, your Identity Provider becomes the _authoritative source_ of information for your 1Password accounts. With Provisioning enabled, the ability to change the _display name_ and _account status_ are not possible through the 1Password Web UI, and must be done through your Identity Provider. You can, however, continue to issue Account Recovery requests through the 1Password Web UI with Provisioning enabled.
+* E-mail addresses for provisioned users cannot be changed through the Identity Provider, due to the way their e-mail addresses are used as a factor for authentication. Users must change their e-mail manually through the 1Password Web UI.
```
grellyd, line 60: This is not true. It makes it seem like a 2FA token. I would change it to something like:

`... due to their email addresses being a component from which the user's encryption keys are derived. Users must change their e-mail manually through the 1Password Web UI so that their encryption key can be re-created with the new email and their existing items re-encrypted with the new encryption key.`

That is overly wordy, but you get the idea.

grellyd: If you are going into detail here, you may want to refer to the whitepaper here which contains further details: https://1password.com/files/1Password-White-Paper.pdf

PR #106, opened 2020-07-09
Add section to Terraform README about region env var

alicethorne-ab:
See: https://github.com/terraform-providers/terraform-provider-aws/issues/9989

If someone doesn't have an `~/.aws/config` file, then this environment variable is necessary. To prevent confusion we should just have them set it to whatever region they want to use.


grellyd:
Perhaps we want to mention the 'if you don't have the aws config file, then...' part. Your phrasing makes it sound non-optional.

PR #110, opened 2020-08-04
Add support for custom resource tags.

jlrgraham23:


grellyd:
Hey @jlrgraham23 

Thanks for the MR. Can you explain a little bit why this is needed?

Our terraform deployment is supposed to be a minimal deployment upon which an admin can expand. Does this addition fit with that minimal vision?

pastjean:
We're coming up with a simpler method to deploy with AWS. This will not be needed anymore. It's gonna be easier to just copy, add your own tags from now own. (this should come up in a week or so)

PR #111, opened 2020-08-04
Add support for specifying an ACM certificate domain.

jlrgraham23:
Allow specification of custom ACM certificates, allows the use of wildcard certs.


grellyd:
Hey @jlrgraham23 

Thanks for the MR. Can you explain a little bit why this is needed?

Our terraform deployment is supposed to be a minimal deployment upon which an admin can expand. Does this addition fit with that minimal vision?

pastjean:
We're coming up with a simpler method to deploy with AWS.
Also we were waiting some feedback we never got.

PR #117, opened 2020-10-13
K8s simplification

pastjean:
The example text is currently long to read and lots of duplication with the Root README.md and PREPARATION.md, this is my take at trying to make the kubernetes deployment simpler/shorter/more concise.

- We encourage to use the provided redis instance here, explicit options to configure the redis host and port are added if you want to use another instace.
- We highly encourage let's encrypt. Let's document it a little, documentation on how to not use it is in /PREPARATION.md, so I remove the "option". An admin can figure it out
- Add a LivelinessProbe to the SCIM bridge deployment. Helps reliability
- Rename the redis deployment to something SCIM Bridge specific
- Fix some typo here and there


kubernetes/op-scim-deployment.yaml
```diff
@@ -16,9 +16,20 @@ spec:
       - name: op-scim
         image: 1password/scim:v1.6.0
         command: ["/op-scim/op-scim"]
-        args: ["--session=/secret/scimsession", "--letsencrypt-domain={YOUR-DOMAIN-HERE}"]
+        args:
+        - "--session=/secret/scimsession"
+        - "--letsencrypt-domain={YOUR-DOMAIN-HERE}"
+        - "--redis-host=redis
```
grellyd, line 22: I assume you are adding these default redis config args as a form of documentation?

kubernetes/op-scim-service.yaml
```diff
@@ -8,12 +8,14 @@ spec:
   type: LoadBalancer
   ports:
   - protocol: TCP
-    name: tls
+    name: https
     port: 443
     targetPort: 8443
+# Traffic on :80 is needed to perform the LetsEncrypt certificate challenges,
```
grellyd, line 14: Also used for the status page if we care to mention that.

grellyd (approved):
LGTM!

One question and a comment. Could be merged as is. Pre-approving.

kubernetes/op-scim-deployment.yaml
pastjean: exactly

kubernetes/op-scim-service.yaml
pastjean: ahh yes I'll add this

PR #120, opened 2020-10-23
Fix issues with AKS deployment

### LINKED ISSUE #119: issue with transition to kubernetes/op-scim-config.yaml
Had a deployment work fine last week to azure kube service before the latest commit. Tried to same deployment today to AKS with the updated code and nothing seems to work. Not listening on 443/80, cert didn't get created etc.. 

I think there's an issue with the vars in the new config.yaml that's being used. Looking into it today will update what I find.

alicethorne-ab:
The `livenessProbe` seems to cause more issues than it solves, so I've removed it for now.

Also fixed a couple of inconsistencies.

Closes #119

### DIFF
--- a/kubernetes/op-scim-deployment.yaml
+++ b/kubernetes/op-scim-deployment.yaml
@@ -13,19 +13,11 @@ spec:
         app: op-scim-bridge
     spec:
       containers:
-      - name: op-scim
+      - name: op-scim-bridge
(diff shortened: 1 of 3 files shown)


pastjean (approved):

platima:
TBH it appears to be getting the wrong IP =/

PR #123, opened 2020-11-17
Introducing AWS Elastic Beanstalk

pastjean:
Elastic beanstalk is a simple PaaS that manages EC2 instances and load-balancers. This will ease your deployments of SCIM bridges as it supports Docker.


aws-elasticbeanstalk-terraform/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+This is only an example of how you could deploy the 1Password SCIM Bridge to your existing AWS infrastructure. As an advanced form of deployment, this AWS/Terraform example can be integrated into your existing infrastructure. Feel free to modify the deployment to suit your specific needs.
+
+# What are we using
+
+We highly recommend using terraform to create this AWS infrastructure, this help reduce the errors while deploying and maintining AWS infrastructure.
+
+# How to deploy the bridge
+
+1. Have the prerequisite "scimsession" file ready, and have read the [preparation document](../Preparation.md)
```
accraw, line 11: This link seems to be broken, at least in the GitHub UI (is it case sensitive?)

aws-elasticbeanstalk-terraform/input.tf
```diff
@@ -0,0 +1,4 @@
+locals  {
+    OP_LETSENCRYPT_DOMAIN = "1password-scimbridge-test.us-east-1.elasticbeanstalk.com"
```
accraw, line 2: It would be nice to have a few details here (for both variables) - does this have to match anything?

aws-elasticbeanstalk-terraform/README.md
accraw: Also, does it mention anywhere that I have to put the scimsession file somewhere in particular?

pastjean: That's exactly why I wanted some help.
I'll fix according to your comments.

- Fix link
- add info on where to put the scimsession

alicethorne-ab:
Looked at this very briefly so far. Couple of things:

1. The docs need to be expanded upon. Perhaps we could take what was already written for AWS/Terraform, and edit them to be more applicable to Beanstalk.
2. Is there any chance we could continue to use SecretsManager for the `scimsession` file? That's what we did previously and it's the more AWS-ic way of doing it.

pastjean:
> Is there any chance we could continue to use SecretsManager for the scimsession file? That's what we did previously and it's the more AWS-ic way of doing it.

SecretsManager is a pretty new AWS service which not much companies use, it would be possible to still use it. It's complicated to understand for a user that has no experience with it, and a user that has experience will know how to set it up.  My main reason of not using it is to keep the example to a bare minimum, when you instroduce "SSM", the IAM permission management starts getting pretty complicated.

PR #125, opened 2020-11-19
Updated to remove ENTRYPOINT command because it's in the image

accraw:
Do not merge until associated ticket is released.


kubernetes/op-scim-deployment.yaml
```diff
@@ -15,7 +15,6 @@ spec:
       containers:
       - name: op-scim-bridge
         image: 1password/scim:v1.6.0
```
pastjean, line 17: this will come at the same time as the new version

pastjean (approved):

PR #130, opened 2021-01-21
Fix targetPort to containerPort deployment

choutone:
As the deployment uses the containerPort `3002`, this PR fixes the `targetPort` service.

### DIFF
--- a/kubernetes/op-scim-service.yaml
+++ b/kubernetes/op-scim-service.yaml
@@ -13,10 +13,10 @@ spec:
   - protocol: TCP
     name: https
     port: 443
-    targetPort: 8443
+    targetPort: 3002
   - protocol: TCP
     name: http
     port: 80
-    targetPort: 8080
+    targetPort: 3002
   selector:
     app: op-scim-bridge


alicethorne-ab:
Hello. Thank you for submitting your PR.

The service typically runs on port `443` and `80` when `OP_LETSENCRYPT_DOMAIN` is specified either in `scim.env` (for Docker) or `op-scim-config.yaml` (for Kubernetes). The SCIM Bridge only runs on `3002` when that variable is not set.

Port `3002` is intended for users who wish to run the SCIM Bridge on custom infrastructure, such as with their own load balancer solution (such as `nginx` or a network-based LB).

Please let me know if you have any other questions about the configuration.

afirth:
for kubernetes you should probably serve on 443 anyway, and simply use the default certificate instead of getting one for letsencrypt. Then AWS ALB and GCP LBs will work like they do with everything else. In the ALB case, it ignores the validity of the cert, and in GCP I believe it trusts the cluster's default cert.

Without either adding 3002 as a port here, or serving on 443 regardless of the setting of letsencrypt domain, this has to be patched by the customer.'

Nobody on K8s professionally wants your service getting its own certs from letsencrypt

BTW, is the scim bridge itself open source?

Minor flaws aside, thanks for shipping this :)

alicethorne-ab:
Thank you for the feedback.

In our experience, it's actually quite common for users of the SCIM bridge to rely on the LetsEncrypt service for their TLS certificate needs. Additionally, relying on cloud providers for SSL can often incur an additional cost, which we're hoping to keep low as possible for this bridge.

We'll definitely take a look at making the Kubernetes example clearer for when users wish to use a custom load balancer, though.

Additionally, the SCIM bridge is not currently open source, and we do not have a roadmap to release it as such at this time.

For now, I'll close this PR, as the changes would not work with the currently-existing instructions or behaviour. Feel free to open an Issue on this repository, though, if you'd like to continue this conversation. Alternatively, you can also reach out to 1Password Business Support.

PR #133, opened 2021-02-12
switch to using required_providers

orenmazor:


aws-ecsfargate-terraform/terraform.tf
```diff
@@ -1,6 +1,13 @@
+terraform {
+  required_version = ">= 0.14.0"
+
+  required_providers {
+    aws = "~> 3.0"
```
perobertson, line 5: The way to specify versions changed recently. This should now be

```
aws = {
  source = "hashicorp/aws"
  version = "~> 3.0"
}
```

orenmazor: coolbeans. applied

PR #147, opened 2021-07-05
Bump to v2.1.0.

alicethorne-ab:
Bumping to v2.1.0: https://app-updates.agilebits.com/product_history/SCIM#v201001

Also adds `depends_on` to Docker Swarm to fix a minor issue a few Swarm users have run into. It's been tested and works.

### DIFF
--- a/docker/swarm/docker-compose.yml
+++ b/docker/swarm/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.3"
 services:
   scim:
-    image: 1password/scim:v2.0.2
+    image: 1password/scim:v2.1.0
     deploy:
       replicas: 1
       restart_policy:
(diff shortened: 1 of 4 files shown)


docker/swarm/docker-compose.yml
```diff
@@ -18,6 +18,7 @@ services:
         gid: "999"
         mode: 0440
     env_file: scim.env
+    depends_on: [ 'redis' ]
```
devillexio, line 21: Is this something we intended to include in this version bump PR? Seems reasonable, but I thought I would double check.

devillexio (approved):
Looks good to me! :tada: One minor comment, but not a blocker.

docker/swarm/docker-compose.yml
alicethorne-ab: I thought I'd include it as it's a minor fix, and I didn't want to create a full PR for a one-line change. It can be broken out if needed, though.

PR #148, opened 2021-07-07
`restart: always` in Docker Swarm example.

alicethorne-ab:
Fixes another issue with Docker Swarm not restarting containers on reboot.

### DIFF
--- a/docker/swarm/docker-compose.yml
+++ b/docker/swarm/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     deploy:
       replicas: 1
       restart_policy:
-        condition: on-failure
+        condition: always
     networks:
       - op-scim
     ports:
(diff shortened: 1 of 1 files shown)


accraw (approved):
Looks good!

docker/swarm/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     deploy:
       replicas: 1
       restart_policy:
-        condition: on-failure
+        condition: always
```
devillexio, line 8: Is `always` the correct value or should it be `any`?

I'm not that familiar with Docker Swarm, but it seems to follow the docker-compose file conventions. According to the [docker-compose docs](https://docs.docker.com/compose/compose-file/compose-file-v3/) these are the valid values for `restart_poilcy.condition`:

image

alicethorne-ab: You are right, unfortunately. Apparently I hadn't tested the correct mode here. I'll open a new PR to fix this bug.

devillexio: My apologies for not getting to the review sooner. 😞

PR #150, opened 2021-08-23
Updated targetPort on Service

dalekurt:
1Password SCIM bridge runs on port 3002.


grellyd:
Hey @dalekurt, 

Thanks for the PR.

There are multiple ways to run the SCIM bridge, and the most common is while leveraging the bundled LetsEncrypt server or via getting a TLS cert through your cloud provider. In those cases, the load balancer is not directly receiving traffic on 3002, and instead is receiving traffic on :8080 and :8443. 

Closing as not applicable.

PR #155, opened 2021-09-15
Update README.md

ag-adampike:
Removed manual droplet creation and reorganized that section, cleaned up a few clerical things, and added instructions to add Redis as a component and bind the connection string to the environment variable.


accraw (approved):
Looks good to me!

PR #165, opened 2021-12-29
Update Terraform example README.md

ag-adampike:
A recent PR (https://github.com/1Password/scim-examples/pull/162) fixed deployment failures for _new_ AWS deployments, but the problem persists for anyone updating to 2.3.0 from an earlier version who isn’t aware of the change to the health check path in the Terraform script.

This PR includes the following changes to the `README` for this deployment example:
- I added a section in the update instructions for updating from versions older than 2.3.0 so the health check will pass.
- I also changed `upgrade` to `update` for consistency with our support articles (i.e. [Update the 1Password SCIM bridge](https://support.1password.com/scim-update/)).


chasdl (approved):
Thank you!

PR #168, opened 2022-01-14
Add pod anti-affinity to kubernetes config

### LINKED ISSUE #167: Add pod anti-affinity config for Kubernetes
We should add pod anti-affinity for `op-scim` and `redis` pods to instruct the Kubernetes scheduler to not place these pods on the same nodes. This helps ensure that resource utilization by the one does not starve or affect the other.

devillecodes:
In this PR we add values for pod anti-affinity to the example Kubernetes config.

The purpose of this change is to request that the Kubernetes scheduler ensures that the SCIM bridge pods and the redis pods are not scheduled on the same node. This is done to ensure that resource utilization by either application does not starve or affect the other. See the Kubernetes documentation for more details: Inter-pod affinity and anti-affinity.

Note that this configuration requires that a minimum node count of 2. This is already the requirement for our one-click deployments.

To test I recommend that you deploy this configuration to your favourite cloud provider, and ensure that the pod anti-affinity settings are applied to both SCIM bridge and Redis pods, and to make sure that they are indeed scheduled on different nodes as desired.

For the below test commands ensure that `<existing-namespace>` is replaced or removed (if you want to use the default namespace).

Example apply command:

```
kubectl apply --namespace=<existing-namespace> -f ./kubernetes
```

Example delete command:

```
kubectl delete --namespace=<existing-namespace> -f ./kubernetes
```

Resolves #167.


alicethorne-ab:
This PR looks great on its own, and I even tested it on AKS.

However, there is a slight complication to the deployment of this PR. I believe we'll need to adjust our documentation [here](https://support.1password.com/scim-deploy-azure/#32-configure-your-cluster) to reflect that *two* nodes are required, as well as offer an "upgrade guide" to get people deployed with one node onto two nodes.

Here's an example command on how to scale the nodes from the CLI, using `az`:

```bash
az aks nodepool scale --cluster-name op-scim-demo --name agentpool --node-count 2 --resource-group op-scim-demo
```

devillexio:
Thanks for the review feedback @alicethorne-ab! 🙌 

> Need to figure out if 2 or 3 nodes is necessary to prevent K8s upgrade halting

What this change means is that we are basically saying that we want to increase the minimum node count from 1 to 2 to allow the SCIM bridge and redis to run on separate nodes.

I recommend that the cluster autoscale also be updated to reflect this, specifically the `min-count` parameter. This will allow the cluster to perform a rolling upgrade of the nodes and to move the pods between nodes without needing to incur any downtime.

I.e. something like:

```
az aks nodepool update --enable-cluster-autoscaler --min-count 2 --max-count 5 -g op-scim-demo -n agentpool --cluster-name op-scim-demo
```

> Need to figure out whether you can make anti-affinity less strict so pods can be on the same node if there is no other option

There is a way to do a "soft" requirement according to the [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity), in the form of `preferredDuringSchedulingIgnoredDuringExecution`. The problem is that in practice this did not work consistently. Meaning there was a good chance k8s would schedule both application pods on the same node even with the anti-affinity configured. This makes have anymore than 1 node in the config redundant.

It made more sense for me to guarantee the behaviour by making it a hard requirement, as opposed to making it a recommendation and then it not being followed. Also, both our GCP and DO marketplace deployments have a minimum node count of 2.

Since this is an example repo, folks can opt to remove the anti-affinity config and use a single node, but then there's no guarantee of resource isolation between op-scim and redis. Of course this is also not required if you opt to use an existing redis and override the redis URL parameter.

> Need to update documentation on our main site, as well as provide a command for folks to run on AKS to re-scale their existing cluster for upgrades

Good point!

- [ ] ensure [support documentation](https://support.1password.com/scim-deploy-azure/ is updated

alicethorne-ab:
Alright, sounds good to me. I'm willing to give this a tentative approval, though I'd urge that we get the Azure documentation update lined up and ready to go live internally before we merge this PR, if that's alright with you.

alicethorne-ab (dismissed):

kubernetes/README.md
```diff
@@ -1,6 +1,7 @@
 # Deploying the 1Password SCIM bridge on Kubernetes
 
 This example explains how to deploy the 1Password SCIM bridge on a Kubernetes cluster. It assumes your Kubernetes cluster supports "load balancer" services.
+It also assumes that your cluster has at least 2 nodes to ensure that the SCIM bridge and Redis pods are note scheduled on the same node.
```
devillexio, line 4:
```suggestion
It also assumes that your cluster has at least 2 nodes to ensure that the SCIM bridge and Redis pods are not scheduled on the same node.
```

devillexio:
Sorry for the re-requested review. I resolved the conflict and updated a typo in the readme.

ag-adampike:
@alicethorne-ab @devillexio relevant docs are now published for AKS (see https://support.1password.com/scim-update/#if-youre-using-azure-kubernetes-service), so this should be ready for merging after dealing with the conflict, unless further review is required for any other reason. :-)

ag-adampike:
Closing this per https://github.com/1Password/scim-examples/issues/167#issuecomment-1790946240.

PR #172, opened 2022-02-17
Fix Terraform example security group

ag-adampike:
This PR fixes a deployment error encountered in certain scenarios.

Without explicitly specifying the `vpc_id` for the `aws_security_group.service` resource, this security group does not necessarily get created in the same VPC as the `aws_security_group.alb` security group.

If the security groups are not created in the same VPC, applying the plan errors and halts, because they are expected to be in the same VPC. 😅

Specifying the VPC from the `aws_vpc` data resource resolves the issue. I also added a comment for consistency with the `aws_security_group.alb` resource.

### DIFF
--- a/aws-ecsfargate-terraform/main.tf
+++ b/aws-ecsfargate-terraform/main.tf
@@ -195,6 +195,8 @@ resource "aws_security_group" "alb" {
 }
 
 resource "aws_security_group" "service" {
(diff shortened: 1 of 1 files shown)


devillexio (approved):
Thanks for finding and fixing this issue! 🙌

PR #173, opened 2022-02-25
ht-6658_remove_secret

humnfrocha:
## jira:
* [HT-6658](https://humnai.atlassian.net/browse/HT-6658)
## motivation:
* Reuse 1password's ECS Fargate deployment
## changes:
* Remove secret from code has it is kept in Terraform state
* Create listener on port 80 in the ALB and redirect to 443
* Refactor security groups to reduced open ports
* Create variable to store the ARN of the secret with 1Password scimsession file
* Create variables to store ECS task memory and cpu
* Add default empty map as default of tags variable
* Remove provider
* Remove restrictions on Terraform version

### DIFF
--- a/aws-ecsfargate-terraform/main.tf
+++ b/aws-ecsfargate-terraform/main.tf
@@ -1,6 +1,4 @@
 terraform {
-  required_version = ">= 0.14.0"
-
   required_providers {
     aws = {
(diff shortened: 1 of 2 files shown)


humnfrocha:
Wrong repository.

PR #175, opened 2022-03-25
Update Kubernetes deployment example

ag-adampike:
This PR is to make several small changes to our generic Kubernetes deployment example:
- removes references to port `8080`, which is no longer necessary with our latest Let's Encrypt implementation
- adds names to the ports on the SCIM bridge and Redis containers and a description of their intended purpose
- references the above names for the `targetPort` in the respective Service specs (rather than explict port number)


devillexio (approved):
This looks great. Thanks so much for submitting these changes! 🙌

PR #178, opened 2022-03-28
Fix HTTP port when not using Let's Encrypt

### LINKED ISSUE #177: SCIM bridge starts in setup mode when not using Let's Encrypt
### Overview

The SCIM bridge shows setup mode when not using Let's Encrypt and accessing the bridge via `http:3002`.

### Steps to produce

1. Deploy 1Password SCIM bridge using [our Kubernetes manifest deployment example](https://github.com/1Password/scim-examples/tree/master/kubernetes) in the default configuration using Let's Encrypt and a public load balancer.
2. After the SCIM bridge is deployed (`kubectl get pods`, `kubectl logs <pod-name>` for status), visit the status page in a browser (e.g. `https://scim.example.com`) and confirm that it is functioning as expected with a sign-in screen accepting the bearer token.
3. Edit [the ConfigMap manifest](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-config.yaml) to configure the SCIM bridge to _not_ use Let's Encrypt (set the value for the `OP_LETSENCRYPT_DOMAIN` environment variable key to `""`).
4. Edit [the SCIM bridge Service manifest](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-service.yaml) to forward traffic on port 3002 to the SCIM bridge service, eg.:
```
…
  ports:
  - protocol: TCP
    name: http
    port: 3002
…
```

4. Re-apply the manifests to update the configuration:
  `kubectl apply -f .`
5. Visit the IP address or URL of the SCIM bridge via HTTP (e.g.: `http://scim.example.com:3002`)

devillexio:
### Overview

In this PR we update the HTTP port for the Load Balancer config. We don't need to change it to `3002` since we now use named ports. 

Note that there are some whitespace changes related to fixing the indentation and removing trailing whitespace in comments.

Resolves #177.

### To test

1. Set the the config value (in `kubernetes/op-scim-config.yaml`) `OP_LETSENCRYPT_DOMAIN` to `""` to disable Let's Encrypt
1. Uncomment the following lines in `kubernetes/op-scim-service.yaml`:
    ```
    #- protocol: TCP
    #  name: http
    #  port: 80
    #  targetPort: http
    ``` 
1. Connect `kubectl` to your favourite cluster
1. Apply this configuration to your cluster `kubectl apply -f .` or `kubectl apply -f ./kubernetes`
1. Navigate to the public IP address of the load balancer or configured domain
1. Ensure that you see the login page for the SCIM bridge, and _not_ the setup page

### DIFF
--- a/kubernetes/op-scim-service.yaml
+++ b/kubernetes/op-scim-service.yaml
@@ -17,7 +17,7 @@ spec:
     # for unencrypted traffic redirected from a reverse proxy or load balancer
     #- protocol: TCP
     #  name: http
-    #  port: 3002
+    #  port: 80
     #  targetPort: http
   selector:
     app: op-scim-bridge
--- a/kubernetes/op-scim-service.yaml
+++ b/kubernetes/op-scim-service.yaml
@@ -15,9 +15,9 @@ spec:
       targetPort: https
     # Use this port when terminating TLS in front of the SCIM bridge to listen
     # for unencrypted traffic redirected from a reverse proxy or load balancer
-    #- protocol: TCP
-    #  name: http
-    #  port: 80
-    #  targetPort: http
+    # - protocol: TCP
+    #   name: http
+    #   port: 80
+    #   targetPort: http
   selector:
     app: op-scim-bridge
--- a/kubernetes/op-scim-service.yaml
+++ b/kubernetes/op-scim-service.yaml
@@ -7,17 +7,17 @@ metadata:
 spec:
   type: LoadBalancer
   ports:
-  # Use this port in the standard configuration to terminate TLS at the SCIM
-  # bridge container using Let's Encrypt
-  - protocol: TCP
-    name: https
-    port: 443
-    targetPort: https
-  # Use this port when terminating TLS in front of the SCIM bridge to listen
-  # for unencrypted traffic redirected from a reverse proxy or load balancer
-  #- protocol: TCP
-  #  name: http
-  #  port: 3002
-  #  targetPort: http
+    # Use this port in the standard configuration to terminate TLS at the SCIM
+    # bridge container using Let's Encrypt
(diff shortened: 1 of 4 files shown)


devillexio:
To see the actual change you can [hide whitespace](https://github.com/1Password/scim-examples/pull/178/files?diff=split&w=1).

kubernetes/op-scim-service.yaml
```diff
@@ -7,17 +7,17 @@ metadata:
...
+    # bridge container using Let's Encrypt
+    - protocol: TCP
+      name: https
+      port: 443
+      targetPort: https
+    # Use this port when terminating TLS in front of the SCIM bridge to listen
+    # for unencrypted traffic redirected from a reverse proxy or load balancer
+    # - protocol: TCP
+    #   name: http
+    #   port: 80
```
laughingman-hass, line 20: is it worth making the ports configurable at default to 80?

devillexio: This is the load balancer port designation. So effectively what you are configuring here is what port to use when you access the bridge via the load balancer, i.e. `http://scim.example.com:port`.

It is already configurable in the sense that you can change `containerPort` value on the deployment:
```
            - name: http
              containerPort: 3002
```

But then you also need to ensure you set `OP_PORT` in the config map to ensure we actually listen to that port.

ag-adampike: I had some thoughts about setting that automatically in the Deployment spec based on the envirnoment variable, but I hadn't actually looked into it beyond that.

Perhaps something to consider for future improvement.

ag-adampike (approved):
LGTM! Thanks for the extra nitpicks on the indents. :chefs-kiss:

PR #180, opened 2022-04-08
Set resource requests and limits for application pods

### LINKED ISSUE #166: Kubernetes deployment example: Deployment containers have no resource limits
Noted in the code linter for the Kubernetes plugin for VS code:
> One or more containers do not have resource limits - this could starve other processes

We should investigate and spec out reasonable default resource limits for the SCIM bridge and Redis containers in their respective deployment manifests ([`op-scim-deployment.yaml`](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-deployment.yaml) and [`op-scim-redis.yaml`](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-redis.yaml)).

See also [Resource Management for Pods and Containers [Kubernetes Documentation](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/).

devillexio:
In this PR we set resource requests and limits for the SCIM bridge and Redis application pods.

We also set a max memory limit for Redis before the Redis server will start evicting keys based on the selected policy.

I tested this by deploying it to a k8s cluster and inspecting the deployments.

Resolves #166.

cc @ag-adampike

### DIFF
--- a/kubernetes/op-redis-config.yaml
+++ b/kubernetes/op-redis-config.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: op-redis-configmap
+data:
+  # Set the arguments passed to Redis
+  # - maxmemory configures Redis to use a specified amount of memory for the data set
+  # - maxmemory-policy is the eviction policy used when the maximum memory is reached
+  REDIS_ARGS: "--maxmemory 256mb --maxmemory-policy volatile-lru"
--- a/kubernetes/op-scim-deployment.yaml
+++ b/kubernetes/op-scim-deployment.yaml
@@ -26,6 +26,13 @@ spec:
             - name: scimsession
               mountPath: "/secret"
               readOnly: false
+          resources:
+            requests:
+              cpu: 125m
+              memory: 256M
+            limits:
+              cpu: 250m
+              memory: 512M
           envFrom:
             - configMapRef:
                 name: op-scim-configmap
--- a/kubernetes/redis-deployment.yaml
+++ b/kubernetes/redis-deployment.yaml
@@ -25,6 +25,6 @@ spec:
             limits:
               cpu: 250m
               memory: 512M
-          env:
-            - name: "REDIS_ARGS"
-              value: "--maxmemory 256mb --maxmemory-policy volatile-lru"
+          envFrom:
+            - configMapRef:
+                name: op-redis-configmap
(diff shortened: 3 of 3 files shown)


laughingman-hass (dismissed):

ag-adampike (commented):
My only question is whether this can or should be set in [the ConfigMap manifest](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-config.yaml). 🤔

ag-adampike:
One more, actually 😅:

Will this close https://github.com/1Password/scim-examples/issues/166? Should we keep that as a "meta" issue for the other deployment examples, or perhaps even create a new one for that?

ag-adampike:
> My only question is whether this can or should be set in [the ConfigMap manifest](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-config.yaml). 🤔

Sorry, I was attmepting to review [these lines](https://github.com/1Password/scim-examples/pull/180/files/354c0d281a5230dfe2db1cbfa7f2e67af25a3929#diff-2bbb700df5fb0f198e097feb7f411b71f416c7907a207e97bd00066b980684beR28-R30) with this comment, but I'm not GitHubbing very well this late on 
a Friday. 😓

devillexio:
> My only question is whether this can or should be set in [the ConfigMap manifest](https://github.com/1Password/scim-examples/blob/master/kubernetes/op-scim-config.yaml).

Good question. I think the resource requests and limits are okay to be defined by us in the yaml templates direct. I do like the idea of making a config map for redis since this is an environment variable we are introducing. I'm going to make that change.

- [ ] introduce config map for redis

devillexio:
> Will this close https://github.com/1Password/scim-examples/issues/166? Should we keep that as a "meta" issue for the other deployment examples, or perhaps even create a new one for that?

You are right. This should close #166. I will update the description to reflect it.

I do have an internal ticket where I'm tracking the changes for the other deployment options. I'll share a link to the issues.

kubernetes/op-redis-config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: op-redis-configmap
+data:
+  # Set the arguments passed to Redis
+  # - maxmemory configures Redis to use a specified amount of memory for the data set
+  # - maxmemory-policy is the eviction policy used when the maximum memory is reached
```
laughingman-hass, line 8: looking forward to this 😄

PR #181, opened 2022-04-12
Set Docker resource requests and limits

devillexio:
This PR updates the arguments passed to Redis to ensure that we have a max limit set for memory usage and select an eviction policy.

I followed the pattern used for the SCIM bridge application and introduced an environment variable file `redis.env` instead of directly injecting the values into the `YAML`.

Similar to #180 did for Kubernetes.

cc @ag-adampike. 🙏

### DIFF
--- a/docker/compose/docker-compose.yml
+++ b/docker/compose/docker-compose.yml
@@ -14,6 +14,7 @@ services:
     image: redis:latest
     networks:
       - op-scim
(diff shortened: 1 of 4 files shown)


laughingman-hass (approved):

ag-adampike (approved):
LGTM! 🎉

PR #182, opened 2022-04-12
Set AWS resource requests and limits

devillexio:
This PR updates the AWS ECS Fargate configuration in a similar way that #181 and #180 did for Docker and Kubernetes.

We set the Redis available memory to 512 and add the `REDIS_ARGS` environment variable.

We also updated the total requested memory to 1024Mb for the deployment since Redis and the SCIM bridge are each allocated 512Mb.

cc @ag-adampike. 🙏

### DIFF
--- a/aws-ecsfargate-terraform/README.md
+++ b/aws-ecsfargate-terraform/README.md
@@ -136,6 +136,20 @@ Connect to your Identity Provider following [the remainder of our setup guide](h
 
 ## Updating
 
+The process for updating your infrastructure involes a few key steps. Lucky for us most of the heavy lifting is done by the `terraform` CLI.
+
+The update steps are generally as follows:
+
+1. Update any variables and task definitions as necessary
+2. Create a plan for Terraform to apply
+3. Apply the new plan to your infrastrucure
+
+Note that the `terraform` CLI will output the details of the plan in addition to saving it to an output file (`./op-scim.plan`). The plan will contain the steps necessary to bring your deployment in line with the latest configuration depending on the changes that are detected. Feel free to inspect the output to get a better idea of the steps that will be taken.
+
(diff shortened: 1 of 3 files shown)


ag-adampike (approved):
Looks good! We should probably note the changes in [the readme](https://github.com/1Password/scim-examples/blob/master/aws-ecsfargate-terraform/README.md), especially since the task definition block in the main Terraform script will have to be updated, and [our current update instructions](https://github.com/1Password/scim-examples/tree/master/aws-ecsfargate-terraform#updating) only mention the task definition itself.

devillexio:
> Looks good! We should probably note the changes in [the readme](https://github.com/1Password/scim-examples/blob/master/aws-ecsfargate-terraform/README.md), especially since the task definition block in the main Terraform script will have to be updated, and [our current update instructions](https://github.com/1Password/scim-examples/tree/master/aws-ecsfargate-terraform#updating) only mention the task definition itself.

I may have misunderstood, but the new plan contains the necessary changes to update the task definition. During testing the plan it came up with was to recreate the task definition to update the required memory resources.

I did add an update section for the changes. Let me know if this is what you had in mind.

ag-adampike:
Not quite. [The instructions to update the bridge](https://github.com/1Password/scim-examples/tree/master/aws-ecsfargate-terraform#updating) only mention patching the image version of our container. This isn't even a traditional update, as there is no new release. What I'm going for here is a way to let all folks currently using a bridge know how to update the configuration itself to the new spec we defined here.

devillexio:
Ah got you! I'll add a separate section for updating the configuration vs updating only the image tag. Thanks for raising it!

ag-adampike (approved):
This is _way_ better than what I had. Closed my PR in favour of this.

PR #191, opened 2022-06-08
Update Kubernetes deployment example README

ag-adampike:
This PR is to fix an issue where renaming the scimsession file causes the bridge to start in setup mode.

The instructions are amended to explicitly set the `scimsession` key when creating the Secret so the bridge doesn't come up in setup mode if the scimsession file is renamed to something else.

To test, confirm that after renaming the scimsession file to something other than `scimsession`, the SCIM bridge deploys correctly and does not start in setup mode when using the `--from-file=scimsession=[<path>/]<filename>` flag with the new filename.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -45,7 +45,7 @@ cd scim-examples/kubernetes/
 
 The following requires that you’ve completed the initial setup of Automated User Provisioning in your 1Password account. See [our support article](https://support.1password.com/scim/#step-1-prepare-your-1password-account) for more details.
 
-Once complete, create a Kubernetes Secret containing the contents of the `scimsession` file. Using `kubectl`, we can read the `scimsession` file and create the Kubernetes Secret in one command (use `--from-file=scimsession=[<path>/]<filename>` if your scimsession file was renamed and/or saved in another directory):
+Once complete, create a Kubernetes Secret containing the contents of the `scimsession` credentials file. Using `--from-file=[key=]source` when [creating the Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret), we can create the Kubernetes Secret named `scimsession`, specify the `scimsession` key, and set its value as the contents of the `scimsession` file in one command:
 
 ```bash
 kubectl create secret generic scimsession --from-file=scimsession=scimsession
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -45,10 +45,10 @@ cd scim-examples/kubernetes/
 
 The following requires that you’ve completed the initial setup of Automated User Provisioning in your 1Password account. See [our support article](https://support.1password.com/scim/#step-1-prepare-your-1password-account) for more details.
 
-Once complete, create a Kubernetes Secret containing the contents of the `scimsession` credentials file. Using `--from-file=[key=]source` when [creating the Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret), we can create the Kubernetes Secret named `scimsession`, specify the `scimsession` key, and set its value as the contents of the `scimsession` file in one command:
+Once complete, create a Kubernetes Secret containing the contents of the `scimsession` credentials file. Using `--from-file=[key=]source` when [creating the Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret), we can create the Kubernetes Secret named `scimsession`, specify the `scimsession` key, and set its value as the contents of the `scimsession` file in one command (replace `./scimsession` if the file is saved in another directory and/or with a different filename):
 
 ```bash
-kubectl create secret generic scimsession --from-file=scimsession=scimsession
+kubectl create secret generic scimsession --from-file=scimsession=./scimsession
 ```
 
 ## (Optional) Create the Google Workspace Kubernetes secrets
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -41,14 +41,14 @@ You can then browse to the Kubernetes directory:
 cd scim-examples/kubernetes/
 ```
 
-## Create the `scimsession` Kubernetes secret
+## Create the `scimsession` Kubernetes Secret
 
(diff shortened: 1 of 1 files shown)


kubernetes/README.md
```diff
@@ -41,14 +41,14 @@ You can then browse to the Kubernetes directory:
...
 ```
 
-## Create the `scimsession` Kubernetes secret
+## Create the `scimsession` Kubernetes Secret
 
-The following requires that you’ve completed the initial setup of Provisioning in your 1Password Account. [See here](https://support.1password.com/scim/#step-1-prepare-your-1password-account) for more details.
+The following requires that you’ve completed the initial setup of Automated User Provisioning in your 1Password account. See [our support article](https://support.1password.com/scim/#step-1-prepare-your-1password-account) for more details.
 
-Once complete, you must create a Kubernetes secret containing the `scimsession` file. Using `kubectl`, we can read the `scimsession` file and create the Kubernetes secret in one command:
+Once complete, create a Kubernetes Secret containing the contents of the `scimsession` file. Using `kubectl`, we can read the `scimsession` file and create the Kubernetes Secret in one command (use `--from-file=scimsession=[<path>/]<filename>` if your scimsession file was renamed and/or saved in another directory):
```
grellyd, line 48: Without explicit context that we are doing something outside of the norm (following kube, but still irregular) this could be confusing. How about:

```suggestion
Once complete, create a Kubernetes Secret containing the contents of the `scimsession` file. Using `kubectl`, we can read the `scimsession` file and create the Kubernetes Secret in one command.

Note this is using the `--from-file=[key=]source` convention of [creating a Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret) to keep the ingested filename consistent in the face of non-scimsession filenames. For example, if your scimsession had been renamed to `1psession`, use the parameter of `--from-file=scimsession=1psession`. 
```

Alternatively we could go with the main case, which is that the scimsession will be named `scimsession` and highlight the key option in the comments. This would require changing the below example to not have scimsession stutter. 

EG:

```suggestion
Once complete, create a Kubernetes Secret containing the contents of the `scimsession` file. Using `kubectl`, we can read the `scimsession` file and create the Kubernetes Secret in one command.

Note this is using the `--from-file=[key=]source` convention of [creating a Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret). If your  file has been renamed, set the optional key parameter to `scimsession` to keep the ingested filename consistent in the face of non-scimsession filenames. For example, if your scimsession had been renamed to `1psession`, use the parameter of `--from-file=scimsession=1psession`. 
```

grellyd (changes requested):
Looking good, but I think we can improve the clarity around what we are doing with the create key value.

kubernetes/README.md
ag-adampike: Thanks @grellyd!

I see where you're going with it, but I don't see this as outside the norm, _per se_. It's a parameter for a command-line operation that allows a key to be specified when creating the Secret object, and SCIM bridge is expecting a key exactly named `scimsession`. It makes sense that we should be instructing customers to use everything up to `kubectl create secret generic scimsession --from-file=scimsession=` at they very least.

The filename itself is very likely to get renamed, especially if we don't call it out at the time of saving, where it might get missed or misunderstood, and it shouldn't be as important as it currently is in any case.

The ideal scenario is the customer follows the readme and saves their scimsession file named `scimsession` in the current directory. We could also add a `./` in front of the filename to make it clearer that this is a path and imply saving it as is in the current directory. I love your explanation, but I think we can sum it all up a bit more concisely.

Perhaps something like this (along with the insertion of `./` in the command below):

```suggestion
Once complete, create a Kubernetes Secret containing the contents of the `scimsession` credentials file. Using the `--from-file=[key=]source` when [creating the Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret), we can specify the `scimsession` key, read the contents of the `scimsession` file, and create the Kubernetes Secret in one command:
```

kubernetes/README.md
```diff
@@ -41,14 +41,14 @@ You can then browse to the Kubernetes directory:
...
 ```
 
-## Create the `scimsession` Kubernetes secret
+## Create the `scimsession` Kubernetes Secret
 
-The following requires that you’ve completed the initial setup of Provisioning in your 1Password Account. [See here](https://support.1password.com/scim/#step-1-prepare-your-1password-account) for more details.
+The following requires that you’ve completed the initial setup of Automated User Provisioning in your 1Password account. See [our support article](https://support.1password.com/scim/#step-1-prepare-your-1password-account) for more details.
 
-Once complete, you must create a Kubernetes secret containing the `scimsession` file. Using `kubectl`, we can read the `scimsession` file and create the Kubernetes secret in one command:
+Once complete, create a Kubernetes Secret containing the contents of the `scimsession` credentials file. Using `--from-file=[key=]source` when [creating the Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret), we can create the Kubernetes Secret named `scimsession`, specify the `scimsession` key, and set its value as the contents of the `scimsession` file in one command:
```
ag-adampike, line 48:
```suggestion
Once complete, create a Kubernetes Secret containing the contents of the `scimsession` credentials file. Using `--from-file=[key=]source` when [creating the Kubernetes Secret](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#create-a-secret), we can create the Kubernetes Secret named `scimsession`, specify the `scimsession` key, and set its value as the contents of the `scimsession` file in one command (replace `./scimsession` if the file is saved in another directory and/or with a different filename):
```

kubernetes/README.md
```diff
@@ -41,14 +41,14 @@ You can then browse to the Kubernetes directory:
...
 
 ```bash
-kubectl create secret generic scimsession --from-file=/path/to/scimsession
+kubectl create secret generic scimsession --from-file=scimsession=scimsession
```
ag-adampike, line 51:
```suggestion
kubectl create secret generic scimsession --from-file=scimsession=./scimsession
```

grellyd (approved):
Excellent. Thanks Adam!

laughingman-hass (approved):
Looking good.... Thank you for this.

PR #193, opened 2022-06-18
Kubernetes deployment example improvements

ag-adampike:
This is a quick PR to rework our Kubernetes deployment example:

- Changes the order of the instructions slightly.
- Defaults to starting without Let's Encrypt by setting the `OP_LETSENCRYPT_DOMAIN` env var to `""`.
- Sets it using `kubectl` _after_ deploying SCIM bridge, allowing the DNS record to propagate before setting it up. Kubernetes does a rolling restart of the pod with the command, so no need for scaling or extra commands, and Let's Encrypt is presumably less likely to fail.
- Revamps the update instructions similarly and for similar reasons; also avoids overwriting config through a pull (note we will want to add this line to the list of items needing to be updated whenever we release a new version, wherever that is tracked).
- Minor clerical fixes throughout sections I was changing: dropping the definite article when referring to SCIM bridge, tweaks for concisencess, readability and clarity.

To test, deploy and test SCIM bridge on a Kubernetes cluster using the updated instructions. Change the image of SCIM bridge to a different version using the update command, and confirm the bridge redeploys using the new version.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -113,7 +113,7 @@ You can now continue with the administration guide to configure your Identity Pr
 
 ## Updating
 
-To update SCIM bridge, connect to your Kubernetes cluster and run the follwing command:
+To update SCIM bridge, connect to your Kubernetes cluster and run the following command:
 
 ```bash
 kubectl set image deploy/op-scim-bridge op-scim-bridge=1password/scim:v2.4.1
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -116,7 +116,7 @@ You can now continue with the administration guide to configure your Identity Pr
 To update SCIM bridge, connect to your Kubernetes cluster and run the following command:
 
 ```bash
-kubectl set image deploy/op-scim-bridge op-scim-bridge=1password/scim:v2.4.1
+kubectl set image deploy/op-scim-bridge op-scim-bridge=1password/scim:v2.5.0
 ```
 
 This will upgrade your SCIM bridge to the latest version, which should take about 2-3 minutes for Kubernetes to process.
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -8,20 +8,20 @@ If you are deploying to the Azure Kubernetes Service, you can refer to our [deta
 
 The deployment process consists of these steps:
 
-1. Create a Kubernetes secret with your `scimsession` file
-1. (Optional) Create Kubernetes secrets for Google Workspace
-2. Configure the SCIM bridge (through the `op-scim-config.yaml` file)
-3. Deploy the SCIM bridge and Redis services
-4. Set up DNS entries for your SCIM bridge
+1. Create a Kubernetes Secret with your `scimsession` file
+2. (Optional) Create Kubernetes Secrets for Google Workspace
+3. Deploy 1Password SCIM bridge to your cluster
+4. Create a DNS record for SCIM bridge
+5. Configure Let's Encrypt
 
 ## Structure
 
-- `op-scim-deployment.yaml`: The SCIM bridge deployment.
-- `op-scim-service.yaml`: Public load balancer for the SCIM bridge server.
-- `op-scim-config.yaml`: Configuration for the SCIM bridge server. You’ll be modifying this file during the deployment.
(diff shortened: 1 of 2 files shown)


kubernetes/README.md
```diff
@@ -107,15 +111,12 @@ curl --header "Authorization: Bearer TOKEN_GOES_HERE" https://<domain>/scim/User
 
 You can now continue with the administration guide to configure your Identity Provider to enable provisioning with your SCIM bridge.
 
-## Upgrading
+## Updating
 
-To upgrade your SCIM bridge, follow these steps:
+To update SCIM bridge, connect to your Kubernetes cluster and run the follwing command:
```
grellyd, line 116:
```suggestion
To update the SCIM bridge, connect to your Kubernetes cluster and run the following command:
```

kubernetes/README.md
```diff
@@ -67,37 +67,41 @@ Then, you will need to provide the Google Service Account key file, also generat
...
 
-At this stage, you can finish configuring your DNS entry as outlined in [PREPARATION.md](/PREPARATION.md).
+Create a public DNS record pointing to this address as outlined in [the preparation guide](/PREPARATION.md).
+
+## Configure Let's Encrypt
+
+After the DNS record above has propagated, run the following command to set the `OP_LETSENCRYPT_DOMAIN` environment variable to the fully-qualified domain name (FQDN) for SCIM bridge based on this record (replace `scim.example.com` with the FQDN):
+
+```bash
+kubectl set env deploy/op-scim-bridge OP_LETSENCRYPT_DOMAIN=scim.example.com
```
grellyd, line 97: Question: given we are no longer setting this value statically into the config, does this persist past redeployments? 

I'll validate this in my testing, but in case you know immediately offhand.

grellyd (approved):
Looks good! I'll give it a test too~

kubernetes/README.md
ag-adampike: It does indeed! 👍

kubernetes/README.md
ag-adampike: This is on purpose (dropping the definite article) and in line with [our style guide](https://support.1password.com/style-guide/#1password-scim-bridge) for referring to product names. 😃

grellyd: Good callout. However `following` is misspelt.

kubernetes/README.md
```diff
@@ -107,15 +111,12 @@ curl --header "Authorization: Bearer TOKEN_GOES_HERE" https://<domain>/scim/User
...
 
-To upgrade your SCIM bridge, follow these steps:
+To update SCIM bridge, connect to your Kubernetes cluster and run the following command:
 
 ```bash
-cd scim-examples/
-git pull
-cd kubernetes/
-kubectl apply -f .
+kubectl set image deploy/op-scim-bridge op-scim-bridge=1password/scim:v2.4.1
```
alicethorne-ab, line 119:
```suggestion
kubectl set image deploy/op-scim-bridge op-scim-bridge=1password/scim:v2.5.0
```

alicethorne-ab (dismissed):
Aside from my one small comment, it looks good to me. ✅

grellyd (approved):
Tested it, and it all works from my end. 👍 

Perhaps however you can tell me a bit more about the rationale behind changing the order of operations? Why not statically declare the domain, instead having to specify it by hand and have the bridge restart?

ag-adampike:
It has to be specified by hand _somewhere_ along the way, either by editing a config file, or running a single Kubernetes command. In either case, Kubernetes persists the configuration to the resulting Pods via the declaration in the Deployment object.

Deploying in this order allows the load balancer time to deploy in a healthy state. After the load balancer has a public address (which can take several minutes), it can be used to create the required DNS record for Let's Encrypt _before_ enabling it on our end by rebooting the container in the correct configuration in one step, rather than mucking about in a text editor at all. It also provides time for the DNS record to propagate after creating it, which sometimes can also take several minutes.

ag-adampike:
To clarify a point in my last comment, a load balancer generally requires a healthy target application, which a SCIM bridge with our Let's Encrypt implementation disabled provides until such time as the required DNS record gets created.

grellyd (approved):
👍 ✅

PR #203, opened 2022-09-13
Remove unused field in workspace settings example

chasdl:
We recently removed the need for consuming the customer's `domain` as part of the workspace configuration set up. This is a bit of cleanup from that work, removing it from an example.


grellyd (approved):
LGTM! ✅ Thanks Chas.

PR #211, opened 2022-09-27
Fix Google Workspace configuration in Docker deployment examples

black-bryan:
This PR cleans up the handling of Docker Secrets when either automatic or manual Docker Swarm deployments are used. Previously the Docker secrets were created but never used by the swarm deployment. Now the Swarm deployment will utilize a secondary `gw-docker-compose.yml` file to read in the created secrets. Also this PR includes cleanup of Docker documentation.

- Workspace secrets are only applied if the secondary `gw-docker-compose.yaml` is added when run through the automatic swarm deployment using [./docker/deploy.sh](https://github.com/1Password/scim-examples/blob/master/docker/deploy.sh) or by specifying this file as a secondary configuration file in the `docker stack deploy` command.
- Removed all references in the docker deployments to Google beta term and adjusted documentation to reflect these changes being available. 
- Altered the structure of the README.md to make the instructions more clear between automatic versus manual deployments. Including the flow of always showing the Docker Swarm deployment option first, since it is the recommended deployment method.

Tested automated and manual deployments of both Swarm deployments with GW as the IdP and non-GW as the IdP. Tested Compose deployments as well.


docker/README.md
```diff
@@ -66,6 +75,52 @@ You can then browse to the Docker directory:
 cd scim-examples/docker/
 ```
 
+#### Docker Swarm
+
+To use Docker Swarm to deploy, you’ll want to have run `docker swarm init` or `docker swarm join` on the target node and completed that portion of the setup. Refer to Docker’s documentation for more details.
```
scottisloud, line 80: Should we link out to the relevant docker documentation here?

docker/deploy.sh
```diff
@@ -78,6 +86,13 @@ run_docker_swarm() {
...
+        if ! docker stack deploy -c $docker_file -c $gw_docker_file op-scim
+        then
+            echo " "
+            echo "Failed to deploy to Docker Swarm; investigate the error before proceeding"
+            sleep 1
+            exit 1
+        fi
     fi
 
     if ! docker stack deploy -c $docker_file op-scim
```
scottisloud, line 98: I wonder if this is necessary? Is this not covered by the `docker stack deploy` on line 67?

docker/deploy.sh
```diff
@@ -78,6 +86,13 @@ run_docker_swarm() {
             sleep 1
             exit 1
         fi
+        if ! docker stack deploy -c $docker_file -c $gw_docker_file op-scim
```
scottisloud, line 89: I wonder if this is necessary? Is this not covered by the `docker stack deploy` on line 67?

docker/deploy.sh
scottisloud, line 98: I wonder if this is necessary? Is this not covered by the `docker stack deploy` on line 67? Looks like it _may_ be a holdover from before your refactor.

docker/README.md
```diff
@@ -66,6 +75,52 @@ You can then browse to the Docker directory:
...
+Next, to create the necessary secrets for Google Workspace:
+
+```bash
+# this is the path of the JSON file you edited in the paragraph above
+cat /path/to/workspace-settings.json | docker secret create workspace-settings -
+# replace <google keyfile> with the name of the file Google generated for your Google Service Account
+cat /path/to/<google keyfile>.json | docker secret create workspace-credentials -
+
+```
+</details>
```
scottisloud, line 101: Do we want to promote this to an H3 heading so it's not tucked behind a disclosure triangle? This would align with my changes to the K8s readme (which doesn't mean my way was right... maybe this is the better way, but we may want to unify on one or the other!

docker/deploy.sh
black-bryan: Line 67 applies the standard docker-compose.yml from the swarm directory if the deployer had selected no for using Google Workspace, whereas line 89 would apply if the deployer had selected yes for using Google Workspace and apply both the docker-compose.yml and the gw-docker-compose.yml from the swarm directory with the settings from gw-docker-compose.yml ontop of the docker-compose.yml.

scottisloud: interesting, my comment may not have attached itself to the correct line. I was thinking of the `if` conditional beginning on line 98. The `if ! workspaceIDP` on line 65 takes care of both conditions, now? If not workspace, deploy using the docker-compose, otherwise deploy using workspace-docker-compose. So that conditional takes care of all possible conditions. 

But line 98 is `another docker stack deploy -c $docker_file` that is not part of any other conditional. Isn't this already covered by TRUE condition of line 65? (If NOT workspace)

docker/deploy.sh
```diff
@@ -196,8 +211,11 @@ done
 # place the domain name into the deployment file, in a backup
 docker_file_path=$docker_path/$docker_type
 docker_file=$docker_file_path/docker-compose.yml
+gw_docker_file=$docker_file_path/gw-docker-compose.yml
 docker_backup_file=$docker_file_path/docker-compose.yml.bak
+gw_docker_backup_file=$docker_file_path/gw_docker-compose.yml.bak
```
scottisloud, line 216: If `$docker_type` is set to "compose", this should, in theory, result in a path like `/scim-examples/docker/compose/gw-docker-compose.yml` but there is no `gw-docker-compose.yml` file in `/scim-examples/docker/compose`. I suspect it's just a matter of creating copies of those yaml files in /compose? (maybe there's another way we can do this without creating duplicate files?)

docker/deploy.sh
black-bryan: Correct, on line 98 it is a holdover from before, I will remove it. Nice catch!

docker/deploy.sh
black-bryan: I have an idea for how to handle this situation cleanly.

docker/README.md
black-bryan: I like the clarity on the headings on the K8 readme, I will transpose this formatting style.

docker/deploy.sh
scottisloud: NICE! 🤩

ag-adampike (approved):
Tested and comnfirmed working as expected in all scenarios (with/without GW, scripted/manual).

Great work @black-bryan!

scottisloud (approved):
Look good, MERGE!

PR #215, opened 2022-10-27
Introduce beta Docker deployment

### LINKED ISSUE #163: `docker-machine` is deprecated
The `docker-machine` command is deprecated, as seen here: https://github.com/docker/machine

### LINKED ISSUE #174: Fix startup order when deploying with Docker Compose
Redis can take some time to start up. The problem is that the SCIM bridge startup fails if it cannot establish a connection to Redis. We should ensure that we give Redis enough time to start up before we attempt to launch the SCIM bridge.

On kubernetes the SCIM bridge will automatically be restarted on a failure, but we don't have that luxury with our Docker-based deployments.

Also, the `depends_on` option we currently rely on is not sufficient. Luckily the documentation describes some potential approaches to ensure the containers are started in the correct order, and with sufficient time for the applications to start up.

See [Control startup and shutdown order in Compose](https://docs.docker.com/compose/startup-order/) for more details.

### LINKED ISSUE #213: Docker Compose v1 is deprecated
Docker Compose v2 graduated to GA earlier this year and Docker Compose v1 was deprecated (see [Announcing Compose V2 General Availability - Docker](https://www.docker.com/blog/announcing-compose-v2-general-availability/)).

Our docs and scripts should be updated to use the new `docker compose` subcommand. A formal investigation is still needed, but there doesn't seem to be any breaking changes for our deployments, and through my experience deploying with customers, simply removing the hyphen to use v2 (i.e. replace all instances of `docker-compose` with `docker compose`) does the trick using Docker Engine without separately installing Docker Desktop or the Docker Compose plugin.

See also [Compose command compatibility with docker-compose | Docker Documentation](https://docs.docker.com/compose/cli-command-compatibility/).

### LINKED ISSUE #214: Docker Compose file versioning is deprecated
The Docker Compose file format is now defined by the [Compose Specification](https://github.com/compose-spec/compose-spec/blob/master/spec.md), and versioning the Compose file is no longer recommended.

Formal investigation and testing is needed to ensure compatibility, but simply removing the version line from each respective file in our Docker deployment examples should suffice:

https://github.com/1Password/scim-examples/blob/51d6f620840a557531bf40ed5003e3349c1d49d9/docker/compose/docker-compose.yml#L1

https://github.com/1Password/scim-examples/blob/51d6f620840a557531bf40ed5003e3349c1d49d9/docker/swarm/docker-compose.yml#L1

See also:
- [Compose specification | Docker Documentation](https://docs.docker.com/compose/compose-file/)
- [Compose file versions and upgrading | Docker Documentation](https://docs.docker.com/compose/compose-file/compose-versioning/)

ag-adampike:
This PR introduces a new deployment example using Docker Swarm and addresses the following outstanding issues for our Docker deployment examples:
- implements the [Compose Specification](https://docs.docker.com/compose/compose-file/)
  - resolves #214

- uses swarm mode to help control startup order (SCIM bridge container will FTL and restart until Redis is available), simplify deployment
  - resolves #163
  - resolves #174
  - resolves #213

Additional improvements:
- introduces a health check for Redis
- templates the `compose.yaml` file to simplify configuration and allow applying from the command line, obviate need to script or manually overwrite any file
- refactors `redis.env` into a native `redis.conf` file (see [Redis configuration](https://redis.io/docs/manual/config/))
- creates Docker secrets and configs for storing credentials and configuration
- removes all unnecessary port mappings
- introduces `resources` spec to ensure adequate resources on the Docker host

### DIFF
--- a/beta/docker/README.md
+++ b/beta/docker/README.md
@@ -3,8 +3,8 @@
 ## Prerequisites
 
 - Linux server with port 443 open
-- Docker Engine (see [Docker Engine installation overview](https://docs.docker.com/engine/install/#server))
-- Docker Compose (see [Install Docker Compose](https://docs.docker.com/compose/install/))
+- Docker Engine (v20.10.3+) (see [Docker Engine installation overview](https://docs.docker.com/engine/install/#server))
+- Docker Compose (v2.0+) (see [Install Docker Compose](https://docs.docker.com/compose/install/))
 
 > **Note:**
 >
--- a/beta/docker/README.md
+++ b/beta/docker/README.md
@@ -1,0 +1,116 @@
+# Docker Compose v2
+
+## Prerequisites
+
+- Linux server with port 443 open
+- Docker Engine (see [Docker Engine installation overview](https://docs.docker.com/engine/install/#server))
+- Docker Compose (see [Install Docker Compose](https://docs.docker.com/compose/install/))
+
+> **Note:**
+>
+> This deployment example assumes Docker Compose is run locally on the Docker
+> host. Deploying remotely (i.e. using `docker context`) is _not_ supported.
+
+## Get started
+
+1. Create a DNS record that points to your Docker host, e.g. `scim.example.com`.
+
+2. Clone this repository and switch to this directory:
+
+    ```sh
+    git clone https://github.com/1Password/scim-examples.git
+    cd ./scim-examples/beta/docker/compose
+    ```
+
+3. Copy your `scimsesion` file to this working directory, e.g.:
+
+    ```sh
+    cp path/to/scimsession ./scimsession
+    ```
+
+4. Create a canonical configuration (replace `scim.example.com` with the DNS record from Step 1):
+
+    ```sh
+    OP_LETSENCRYPT_DOMAIN=scim.example.com docker compose -f compose.template.yaml convert > compose.yaml
+    ```
+
+### Configure 1Password SCIM bridge to connect to Google Workspace
+
+If integrating 1Password with Google Workspace, additional cofiguration is required. See [Connect Google Workspace to 1Password SCIM Bridge](https://support.1password.com/scim-google-workspace/#step-1-create-a-google-service-account-key-and-api-client) for instructions to create the service account, key, and API client.
+
+1. Copy the Google Workspace service account key to this working directory, i.e.:
+
+    ```sh
+     cp path/to/workspace-credentials.json ./workspace-credentials.json
+     ```
+
+2. Copy the [settings file template](/beta/workspace-settings.json) to this directory:
+
+    ```sh
+    cp ../workspace-settings.json ./workspace-settings.json
+    ```
+
+    Edit the file. Replace the values for each key:
+
+    - `actor`: the email address for the administrator that the service account is acting on behalf of
+    - `bridgeAddress`: the URL for your SCIM bridge based on the DNS record created in [Get started](#get-started)
+
+    ```json
+    {
+        "actor":"admin@example.com",
+        "bridgeAddress":"https://scim.example.com"
+    }
+    ```
+
+    Save the file.
+
+3. Update the canonical configuration for SCIM bridge to include the overlay configuration for Google Workspace (replace `scim.example.com` with the DNS record for your SCIM bridge):
+
+    ```sh
+    OP_LETSENCRYPT_DOMAIN=scim.example.com docker compose -f compose.template.yaml -f compose.gw.yaml convert > compose.yaml
+    ```
+
+## Deploy 1Password SCIM bridge
+
+After the DNS record has propagated, deploy 1Password SCIM bridge using Docker Compose:
+
+```sh
+docker compose up -d
+```
+
+## Appendix
+
+### Update 1Password SCIM bridge
+
+You can monitor the [1Password SCIM bridge release notes](https://app-updates.agilebits.com/product_history/SCIM) page to see the latest version.
+
+To update your SCIM bridge, switch to this directory, pull the latest version of this repository, and apply the updated `.yaml` (replace `scim.example.com` with the DNS record pointing to your SCIM bridge):
+
+```sh
+cd scim-examples/beta/docker
+git pull
+OP_LETSENCRYPT_DOMAIN=scim.example.com docker compose -f compose.template.yaml [-f compose.gw.yaml] convert > compose.yaml
+docker compose up -d
+```
+
+### Use regenerated credentials
+
+If you want to use a new bearer token or `scimsession` file, [visit the Integrations page](https://start.1password.com/integrations/provisioning) at 1Password.com and click Regenerate Credentials. Copy the new `scimsession` file to the Docker host and restart SCIM bridge, e.g.:
+
+```sh
+cd scim-examples/beta/docker
+cp path/to/scimsession ./scimsession
+docker compose restart scim
+```
+
+The bearer token and `scimsession` file are cryptographically linked. Update the configuration in your identity provider to use the new bearer token.
+
+### Change the DNS name
+
+To use a new DNS name for your SCIM bridge, create a new record pointing to your bridge, wait for it to propagate, then redeploy, i.e.:
+
+```sh
+cd scim-examples/beta/docker
+OP_LETSENCRYPT_DOMAIN=new-name.example.com docker compose -f compose.template.yaml [-f compose.gw.yaml] convert > compose.yaml
+docker compose up -d
+```
--- a/beta/docker/README.md
+++ b/beta/docker/README.md
@@ -6,29 +6,30 @@
 - Docker Engine (v20.10.3+) (see [Docker Engine installation overview](https://docs.docker.com/engine/install/#server))
 - Docker Compose (v2.0+) (see [Install Docker Compose](https://docs.docker.com/compose/install/))
 
-> **Note:**
+> **Note**
 >
 > This deployment example assumes Docker Compose is run locally on the Docker
-> host. Deploying remotely (i.e. using `docker context`) is _not_ supported.
+> host. Deploying to a remote Docker host (i.e. using `docker context`) is
+> _not_ supported.
 
 ## Get started
 
 1. Create a DNS record that points to your Docker host, e.g. `scim.example.com`.
 
-2. Clone this repository and switch to this directory:
+2. On the Docker host, clone this repository and switch to this directory:
 
     ```sh
     git clone https://github.com/1Password/scim-examples.git
-    cd ./scim-examples/beta/docker/compose
+    cd ./scim-examples/beta/docker
     ```
 
-3. Copy your `scimsesion` file to this working directory, e.g.:
+3. Copy your `scimsesion` file to the working directory on the Docker host, e.g.:
 
     ```sh
-    cp path/to/scimsession ./scimsession
+    scp local/path/to/scimsession user@host:scim-examples/beta/docker/scimsession
     ```
 
-4. Create a canonical configuration (replace `scim.example.com` with the DNS record from Step 1):
+4. From the working directory on the Docker host, create a canonical configuration (replace `scim.example.com` with the DNS record from Step 1):
 
     ```sh
     OP_LETSENCRYPT_DOMAIN=scim.example.com docker compose -f compose.template.yaml convert > compose.yaml
@@ -38,13 +39,13 @@
 
 If integrating 1Password with Google Workspace, additional cofiguration is required. See [Connect Google Workspace to 1Password SCIM Bridge](https://support.1password.com/scim-google-workspace/#step-1-create-a-google-service-account-key-and-api-client) for instructions to create the service account, key, and API client.
 
-1. Copy the Google Workspace service account key to this working directory, i.e.:
+1. Copy the Google Workspace service account key to this working directory on the Docker host, e.g.:
 
     ```sh
-     cp path/to/workspace-credentials.json ./workspace-credentials.json
-     ```
+     scp local/path/to/workspace-credentials.json user@host:scim-examples/beta/docker/workspace-credentials.json
+    ```
 
-2. Copy the [settings file template](/beta/workspace-settings.json) to this directory:
+2. From the working directory on the Docker host, copy the [settings file template](/beta/workspace-settings.json) from this repository:
 
     ```sh
     cp ../workspace-settings.json ./workspace-settings.json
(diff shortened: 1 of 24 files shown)


beta/docker/README.md
```diff
@@ -0,0 +1,126 @@
...
+    ```sh
+    git clone https://github.com/1Password/scim-examples.git
+    cd ./scim-examples/beta/docker
+    ```
+
+3. Copy your `scimsesion` file to the working directory on the Docker host, e.g.:
+
+    ```sh
+    scp local/path/to/scimsession user@host:scim-examples/beta/docker/scimsession
+    ```
```
scottisloud, line 30: Instruct user to `chmod 644 scimsession` to handle the likely scenario they only had 600,

beta/docker/README.md
```diff
@@ -0,0 +1,126 @@
...
+### Configure 1Password SCIM bridge to connect to Google Workspace
+
+If integrating 1Password with Google Workspace, additional cofiguration is required. See [Connect Google Workspace to 1Password SCIM Bridge](https://support.1password.com/scim-google-workspace/#step-1-create-a-google-service-account-key-and-api-client) for instructions to create the service account, key, and API client.
+
+1. Copy the Google Workspace service account key to this working directory on the Docker host, e.g.:
+
+    ```sh
+     scp local/path/to/workspace-credentials.json user@host:scim-examples/beta/docker/workspace-credentials.json
+    ```
+
```
scottisloud, line 47: Here too, `chmod 644 workspace-credentials.json`

scottisloud (changes requested):
All looks good. Successful deployment based on these instructions and manifest. Permissions issue could be experienced with uploaded files so added comments to document the chmod needed to resolve the issue.

ag-adampike:
Working on some changes to this deployment:

- default to Docker Swarm
- rework the setup accordingly
  - start with `docker swarm init`
  - pipe `config` into `stack deploy` instead of writing to disk
- rewrite README

ag-adampike:
@1Password/1password-solutions-rw have at 'er! Please feel free to merge it in my absence if you feel it's ready! 😘

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+- Docker Engine (v20.10.3+) (see [Docker Engine installation overview](https://docs.docker.com/engine/install/#server))
+- public DNS record pointing to the Linux server
+
+## Get started
+
+1. Create a public DNS record that points to the public IP address of the Linux server for your SCIM bridge. For example, `scim.example.com`.
+
+2. [Install Docker Engine](https://docs.docker.com/engine/install/#server) on the server (follow the Server instructions; Docker Desktop is not needed). We also recommend following the [post-install steps](https://docs.docker.com/engine/install/linux-postinstall/) as noted in the documentation to enable running Docker as a non-root user and ensure that Docker Engine starts when the Linux system boots.
+
+3. All following steps can be run directly on the Linux server, or by connecting remotely using the `--host` parameter or `docker context`. If you are already using SSH to connect to the Linux server, you can create and use a Docker context on your local computer that is connected to your 1Password account (these commands require Docker Engine or Docker Desktop to be installed on your computer).
```
scottisloud, line 29: I wonder if we should be as simple and straightforward as possible here. Maybe we should just lean into the "directly on the Linux server" and if someone is experienced enough and knows how to do this other ways, they can go ahead and do so. 

I'm not exactly sure that is meant by
>connected to your 1Password account

Just a couple thoughts, no demands here.

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+    git clone https://github.com/1Password/scim-examples.git
+    cd ./scim-examples/beta/docker
+    ```
+
+6. Save your `scimsession` file to the working directory.
+
+   > **Note**
+   >
+   > If you are deploying SCIM bridge directly on the Linux server, you will need to transfer the `scimsession` file
+   > from your local machine that is signed in to your 1Password account, for example:
```
scottisloud, line 64: > from your local machine that is signed in to your 1Password account, for example:
Another reference to "signed in to your 1Password account" that I'm not sure I understand the significance of. 

Maybe we need to clarify this at the beginning of the docs so that what this means, and it's significance, can be better understood throughout? 

In theory this _could_ be accomplished by someone who has the scimsession file and isn't yet a member of the account the SCIM bridge is associated with, so theoretically as long as you have the credentials you don't even need to be signed in nor have an account.

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+### Configure 1Password SCIM bridge to connect to Google Workspace
+
+If integrating 1Password with Google Workspace, additional cofiguration is required. See [Connect Google Workspace to 1Password SCIM Bridge](https://support.1password.com/scim-google-workspace/#step-1-create-a-google-service-account-key-and-api-client) for instructions to create the service account, key, and API client.
+
+1. Copy the Google Workspace service account key to the working directory.
+
+   > **Note**
+   >
+   > If you are deploying SCIM bridge directly on the Linux server, you will need to transfer the Workspace credentials file
+   > from your local machine that is signed in to Google (or your 1Password account if you saved the credentials there), for example:
```
scottisloud, line 79: Same as above re: signed into 1Password account

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+## Deploy 1Password SCIM bridge
+
+After the DNS record has propagated, set the domain name for your SCIM bridge inline, use the template to output a canonical configuration that will create a Docker secret and deploy SCIM bridge, then create the stack from this configuration. You SCIM bridge should automatically acquire and manage a TLS certificate from Let's Encrypt on your behlaf using the supplied domain name.
+
+*Example command:*
+
+```sh
+OP_TLS_DOMAIN=scim.example.com docker stack config --compose-file ./compose.template.yaml |
+    docker stack deploy --compose-file - op-scim-bridge
+```
```
plttn, line 114: We're already pulling in `scim.env` as an env-file. We should be having users edit those rather than potentially setting them up for a poor situation where `OP_TLS_DOMAIN` is set multiple places.

beta/docker/compose.gw.yaml
```diff
@@ -0,0 +1,13 @@
+services:
+  scim:
+    secrets:
+      - workspace-credentials
+      - workspace-secrets
+    environment:
+      - OP_WORKSPACE_CREDENTIALS=/run/secrets/workspace-credentials.json
```
plttn, line 7: Being set multiple places. I would lean using `scim.env` for all variables that end users will need to modify.

beta/docker/compose.template.yaml
```diff
@@ -0,0 +1,58 @@
...
+      timeout: 3s
+      retries: 5
+  scim:
+    image: 1password/scim:v2.8.2
+    depends_on:
+      - redis
+    secrets:
+      - scimsession
+    environment:
+      - OP_TLS_DOMAIN
```
plttn, line 32: Use `scim.env` again

plttn (changes requested):
High level thoughts: 

- We mix where changes should belong. Any user changes should either not be in a `template.yaml` or should instead be in `scim.env`.
    - this can be resolved by moving all "hardcoded" values (`OP_SESSION`, `OP_WORKSPACE_*`) to defined env variables in `compose.template.yaml`, and removing them from `scim.env`. `compose.template.yaml` can then be renamed to `compose.yaml` as there's no changes that need to be made to the yaml. 
- Using `.gitignore` to ensure that a end user's `scim.env` file is not uploaded may make sense here as well. Rename `scim.env` to `scim.env.template` and add `scim.env` to the gitignore

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+Replace `scim.example.com` with the fully qualified domain name that points to your Linux server, then run the command to deploy your SCIM bridge.
+
+### Connect your SCIM bridge to Google Workspace
```
scottisloud, line 118: The organization of the docs here may be a source of some confusion for Workspace customers. 

It's conceivable (I'd wager it's _likely_) that a Workspace customer will have already followed the directions at line 112.

This is and admittedly tricky one to navigate from a documentation standpoint, but I think we just need to do something to foreshadow to a Workspace customer, around line 112, that they should NOT run the command at line 112. 

This could either be a specific call-out like "if you are not using Google Workspace...."

Or it could be a matter of putting the two commands visually close to each other so Workspace and non-Workspace customers can easily see that there are two paths and pick the appropriate one. This could be perhaps putting _both_ behind disclosure arrows with titles that they can easily see the two paths and expand the relevant one.

beta/docker/compose.gw.yaml
plttn: Since end users will not be changing the location of workspace credentials, I retract this comment

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+```
+
+You can sign in to the URL of your SCIM bridge based on its fully qualified domain name (for example, <https://scim.example.com/>). Sign in with your bearer token to test the connection, view status information, or retrieve logs.
+
+You can also send an authenticated SCIM API request to your SCIM bridge from the command line.
+
+*Example command:*
+
+```sh
+curl --header "Authorization: Bearer mF_9.B5f-4.1JqM" https://scim.example.com/Users
```
scottisloud, line 148: LOVE THIS

scottisloud, line 148:
```suggestion
curl --header "Authorization: Bearer mF_9.B5f-4.1JqM" https://scim.example.com/scim/Users
```

scottisloud (changes requested):
I am very stoked about these improvements and I think there are so many really good reasons to deprecate Compose altogether and go all-in on a one-node Swarm deployment.

The approach you are taking with the deployment and documentation is ambitious. I think there are some ways that the documentation can be simplified. @plttn has already highlighted some concerns I share about the specific way the configuration files are being handled. 

Taken together I think once you address both sets of concerns this will be a dramatic improvement and modernization to the current Docker offerings.

beta/docker/README.md
ag-adampike: My thoughts here are that someone needs to get the `scimsession` credentials and transfer them to the server where they are deploying their SCIM bridge, and this provides a sensible means to do so.

In the common case where someone is working from their daily driver workstation that they use to set up provisioning  for their 1Password account and using SSH to connect to the Linux server, this provides an easy means to create Docker secrets from the same workstation, and potentially even load them straight from their 1Password account via `op read` the same way I hint at for the bearer token in the callout. I intended to throw some more similar hints about that.

Creating a Docker context also sets it up nicely for the same admin to manage it in the future or as reference for someone else to create to do the same.

beta/docker/README.md
ag-adampike: I tested and planned for this. When you run this command, you should get a new or updated stack for a SCIM bridge that is connected to Google Workspace. If you go through the general instructions on line 112 first, you get a SCIM bridge that's not connected to Google Workspace first until you run this one.

beta/docker/README.md
ag-adampike: There's a method to my madness here based on [order of precedence for environment variables in Docker](https://docs.docker.com/compose/environment-variables/envvars-precedence/), but it could potentially use some tweaking for sure.

I have intentionally been trying to avoid editing files at all, especially for the most common use case. Setting variables inline means you can deploy in one step without editing anything. Providing `scim.env` with default config allows customers to use it as a base for their own customized setup if they want. It's also useful as a reference to hold information about the things that may be configured, either by editing this file or setting values inline.

ag-adampike:
Thanks for the reviews friends! I want to discuss your concerns and consider your suggestions some more when we can chat it out. 😘

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+        "actor":"admin@example.com",
+        "bridgeAddress":"https://scim.example.com"
+    }
+    ```
+
+    Save the file.
+
+## Deploy 1Password SCIM bridge
+
+After the DNS record has propagated, set the domain name for your SCIM bridge inline, use the template to output a canonical configuration that will create a Docker secret and deploy SCIM bridge, then create the stack from this configuration. You SCIM bridge should automatically acquire and manage a TLS certificate from Let's Encrypt on your behlaf using the supplied domain name.
```
black-bryan, line 107: Spelling fix, and suggestion on what is being set. 
```suggestion
After the DNS record has propagated, set the domain name for your SCIM bridge inline, use the template to output a canonical configuration that will create a Docker secret and deploy SCIM bridge, then create the stack from this configuration. You SCIM bridge should automatically acquire and manage a TLS certificate from Let's Encrypt on your behalf using the fully-qualified domain name (FQDN) for SCIM bridge based on this environment variable.
```

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+*Example command:*
+
+```sh
+OP_TLS_DOMAIN=scim.example.com docker stack config --compose-file ./compose.template.yaml |
+    docker stack deploy --compose-file - op-scim-bridge
+```
+
+Replace `scim.example.com` with the fully qualified domain name that points to your Linux server, then run the command to deploy your SCIM bridge.
```
black-bryan, line 116: following same wording/attached as other examples

```suggestion
Replace `scim.example.com` with the fully-qualified domain name (FQDN) that was created in [Get started](#get-started), then run the command to deploy your SCIM bridge.
```

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+*Example command:*
+
+```sh
+OP_TLS_DOMAIN=scim.example.com docker stack config \
+    --compose-file ./compose.template.yaml \
+    --compose-file ./compose.gw.yaml |
+        docker stack deploy --compose-file - op-scim-bridge
+```
+
+Replace `scim.example.com` with the fully qualified domain name that points to your Linux server, then run the command to deploy SCIM bridge with access to the required configuration and credentials.
```
black-bryan, line 131: Same comment as above

beta/docker/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Test your SCIM bridge
+
+Run this command to retrieve logs from the service for the SCIM bridge container:
+
+```sh
+docker service logs op-scim-bridge_scim
+```
+
+You can sign in to the URL of your SCIM bridge based on its fully qualified domain name (for example, <https://scim.example.com/>). Sign in with your bearer token to test the connection, view status information, or retrieve logs.
```
black-bryan, line 141: We should be consistent throughout on how we reference the FQDN or the DNS record in our example wording on what to replace.

black-bryan (commented):
Love how this is coming along. There are a few places were we start to reference Linux domain name of the server vs the FQDN.

beta/docker/README.md
scottisloud: Rock on.  I don't remember where I even got the idea that led me to make this suggestion but trust you 100%

beta/docker/README.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Overview
+
+The open source Docker Engine tooling can be used to deploy 1Password SCIM Bridge on any supported Linux distribution. A single-node swarm is sufficient. This example assumes that the Linux server is exposed directly to the public internet with Docker acting as a reverse proxy to the SCIM bridge container. Other configurations are possible, but not documented in this deployment example.
+
+## Prerequisites
+
+- AMD64 VM or bare metal server with a Docker-supported Linux distribution (e.g. Ubuntu, Debian, Fedora, etc.)
+- Docker Engine (see [Docker Engine installation overview](https://docs.docker.com/engine/install/#server)) installed on the Linux server
+- a public DNS record pointing to the Linux server
```
plttn, line 21: Public DNS A record?

beta/docker/README.md
```diff
@@ -0,0 +1,227 @@
...
+- SSH access to the Linux server
+- [Docker Desktop](https://docs.docker.com/desktop/install/) or Docker Engine installed on a machine that can connect via SSH to the Linux server and to your 1Password account
+
+## Get started
+
+Before proceeding, you should review the [Preparation Guide](/PREPARATION.md) at the root of this repository. Create a public DNS record that points to the public IP address of the Linux server for your SCIM bridge. For example, `scim.example.com`.
```
plttn, line 27: ditto

plttn (changes requested):

plttn (approved):
LGTM

PR #216, opened 2022-11-13
Add Heroku example

ed-flanagan:
👋 Hello, I'd love to help add Heroku support for the SCIM bridge.

* Add a `/heroku` section
    * TODO: flesh the README out more. Can do if want to move forward
* Update the root README to link to the `/heroku` section
* Add an `app.json` with values I believe are pertinent for a Heroku deployment of the SCIM Bridge
    * General app config like description, keywords, logo, etc.
    * A Redis addon with maxmemory policy
    * `OP_SESSION` the only required value, with additional optional vars
* Add a "shim" Dockerfile for Heroku manifest
    * Add `CMD` specific for Heroku use case. We want to reference config values at the `run` stage, since they will not be available during `build`[^1][^3]
    * Heroku needs container applications to use the `$PORT` value they provide[^2]. I.e. cannot use the default `3002`, so `--port` is used
    * I'd like to replace `OP_REDIS_URL` with `OP_REDIS_TLS_URL`, more on that below
    * I originally had this in a shim script, but shouldn't be necessary unless some flags become optional
* Add a `heroku.yml` manifest to build & use the shim Dockerfile
 
---

**Local testing:**

```bash
cd $PROJ_ROOT/heroku
docker build --tag heroku-scim:latest .
docker run -p 3004:3004 -e PORT=3004 -e OP_REDIS_URL="<redis url from heroku>" -e OP_SESSION="<base64-encoded session content>" heroku-scim:latest
```

**Remote testing:**

* Change the deploy button link to point to the branch (since upstream won't have the changes until merged)
  ```patch
  --- a/heroku/README.md
  +++ b/heroku/README.md
  @@ -1,3 +1,3 @@
   # Deploying the 1Password SCIM Bridge on Heroku
  
  -[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com  /deploy?template=https://github.com/1Password/scim-examples)
  +[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/ed-flanagan/scim-examples/tree/add-heroku)
  ```
* Fill out deployment page like normal

---

**Heroku Redis TLS issues**

When using the `heroku-redis` addon, you're provided with a `_URL` and `_TLS_URL` config variable.
When using the TLS URL, the application gives an error similar to this:

```
FTL failed to set up redis for SCIM bridge error="Network: (failed to NewRedisConnPool), cannot initialize connection: cannot connect to redis: x509: certificate is valid for <SUBDOMAIN>.service.shogun.heroku.com, not <EC2-ID>.compute-<id>.amazonaws.com" application=op-scim build=206022 function=GetCache version=2.6.2
```

Unfortunately, Heroku requires you _skip certificate verification_[^4]. To use. Seemingly this is because the certs are self-signed or something.
 
Best I can tell, `op-scim` is a Go binary built with v1.18.5. Heroku provides an [example on disabling this verification](https://devcenter.heroku.com/articles/connecting-heroku-redis#connecting-in-go).
I couldn't really tell what library y'all use, but regardless of client, I believe `InsecureSkipVerify` in [tls.Config](https://pkg.go.dev/crypto/tls#Config) needs to be set to `true`.

I think adding a configurable option, e.g. `TLS_INSECURE` to configure the underlying client could allow resolution for this issue.
I could then update the shim script to adjust the URL based on that value.

[^1]: https://devcenter.heroku.com/articles/container-registry-and-runtime#dockerfile-commands-and-runtime
[^2]: https://devcenter.heroku.com/articles/container-registry-and-runtime#get-the-port-from-the-environment-variable
[^3]: https://devcenter.heroku.com/articles/build-docker-images-heroku-yml#setting-build-time-environment-variables
[^4]: https://devcenter.heroku.com/articles/heroku-redis#security-and-compliance


ed-flanagan:
> a customer get a SCIM bridge up and running on Heroku which ultimately ran into the same issue you've noted re: TLS/redis, at which point the customer opted to deploy on another platform.

Pretty much in the same boat. TLS issue's was a deal-breaker. Ended up using AWS AppRunner. Happy to share that Terraform too, but would need some time to clean up for more generic use.

> Since it looks like the issue may require modifications to the SCIM bridge code itself, I'll also get this in front of those developers to investigate what may be required there.... We'll keep you posted as we make progress on our side.

Amazing, appreciate it

scottisloud:
Hi ed-flanagan, 
After some internal discussion, we determined that we are not currently positioned to provide official support and maintenance of a Heroku deployment of the 1Password SCIM bridge at this time. For that reason, we will not be merging in your PR. That being said, we will make an issue to explore the modifications to the Redis TLS configuration necessary to support Heroku deployments in the future 

Thank you again for your PR! We really appreciate you wishing to get involved.

ed-flanagan:
Understood. Mind if I open an issue for the Heroku/Redis TLS thing? Just so I have something to track for any later updates. I'll paste what I have and link back to this PR discussion.

PR #217, opened 2022-11-15
Update kubernetes/README.me to correct typo

plttn:
Fix link error, link references `op-scim-service` but points at `op-scim-config`.


ag-adampike (approved):
Good catch. Thanks for filing this! ♥️

PR #218, opened 2022-11-28
Add note about increasing bridge and redis deployments, add commands

chasdl:
This work stems mainly from https://gitlab.1password.io/dev/b5/op/-/issues/3008.

Speaking with customers, our current documentation is not clear that we are recommending resources be upgraded for both the SCIM bridge _and_ Redis deployments, so we clarify that here.

Also, adding k8s commands directly in the `README.md` can reduce customer service inquires, and adding them explicitly can further clarify that both deployments need upgrading.

The actual values proposed are not changed, but I removed the parenthetical values because they only add confusion.


grellyd (approved):
Thanks for making these changes Chas. Much clearer!

PR #220, opened 2022-12-05
Terraform security group enhancements

ag-adampike:
This PR refreshes the ingress and egress rules for the security groups created by this Terraform module to tighten the security requirements; more specifically:

- removes the redundant rule for the load balancer allowing incoming HTTP traffic to ensure a secured connection from the identity provider
- updates the egress rule for the load balancer to allow only TCP traffic to the VPC for SCIM bridge
- update the egress rule for the ECS service for SCIM bridge to allow only HTTPS traffic

To test:

- [x] deploy SCIM bridge from scratch using the Terraform example from this branch
- [x] deploy SCIM bridge from `main`, switch to this branch, and run `terraform apply`
- [x] confirm provisioning works as expected on a SCIM bridge deployed from this branch

### DIFF
--- a/aws-ecsfargate-terraform/main.tf
+++ b/aws-ecsfargate-terraform/main.tf
@@ -222,7 +222,7 @@ resource "aws_security_group" "service" {
   # Create a security group for the service
   vpc_id = data.aws_vpc.this.id
 
-  # Only allow traffic from the load balancer security group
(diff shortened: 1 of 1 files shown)


black-bryan (approved):
LGTM. 🙌

PR #227, opened 2023-02-22
Updating Kub-recommended resources

Hemal1P:
Modifying the Kubernetes recommended resources for different types of deployments.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -191,15 +191,15 @@ Our current default resource requirements (defined in [op-scim-deployment](https
 <details>
   <summary>Default</summary>
   
-    ```yaml
-    requests:
-      cpu: 125m
-      memory: 256M
+  ```yaml
+  requests:
+    cpu: 125m
+    memory: 256M
 
-    limits:
-      cpu: 250m
-      memory: 512M
-    ```
+  limits:
+    cpu: 250m
+    memory: 512M
+  ```
 </details>
 
  Note that these are the recommended `requests` and `limits` values for both the SCIM bridge and Redis containers. These values can be scaled down again after the high volume deployment.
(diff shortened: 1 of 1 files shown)


chasdl:
Thanks Hemal! I'm thinking we can modify the table values. `2000` is a bit low for the high volume deployment, let's adjust the range a bit to `5000`. Let's do:

| 100 - 5000  |  High Volume Deployment  |
| 5000+  |  Very High Volume Deployment  |

black-bryan (commented):
In [line 181](https://github.com/1Password/scim-examples/blob/e66b8d9b014ba11af3ec3c331ea34828065252e8/kubernetes/README.md?plain=1#L181), we call out **high volume deployments where there is a large number of users and/or groups.** This seems like it should be 1000-5000 for high volume and default being 1-1000 users? Above 100 users does not seem like a large number of users for most accounts.

chasdl:
@ag-adampike Has brought up some research on K8's resource management that we should take the time to explore:

[CPU requests should be defined, but limits should not](https://home.robusta.dev/blog/stop-using-cpu-limits) (to allow for efficient use of available CPU)
[Memory limits should be set to the same value as requests](https://home.robusta.dev/blog/kubernetes-memory-limit) (to avoid OOM kills)

For some historical context, we have been configuring both `requests` and `limits` per Kubernetes documentation ([on CPU](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#if-you-do-not-specify-a-cpu-limit), on [Memory](https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/#if-you-do-not-specify-a-memory-limit)). The reasons outlined above for not setting limits on CPU and setting the limit == request make sense, but we need to do more research if it fits our particular use-case.

chasdl:
@Hemal1P Any luck verifying your commits? You should be able to upload your public key.

Hemal1P:
Hey @chasdl Thank you !  it worked 🙌

chasdl (approved):

PR #230, opened 2023-02-27
Fix FTL in v2.7.4 and use TLS Secret type for custom certs

### LINKED ISSUE #229: Kubernetes deployment fails to load if certificate and key files are not found
When `OP_TLS_DOMAIN` (or `OP_LETSENCRYPT_DOMAIN`), `OP_TLS_CERT_FILE`, and `OP_TLS_KEY_FILE` are set for the `op-scim-bridge` Deployment, the SCIM bridge container fails to load:

```sh
FTL failed to generate TLS config, retrying after backoff delay error="could not find cert file with given path secrets/tls-certificate: stat /secrets/tls-certificate: no such file or directory" application=op-scim build=207041 version=2.7.4
```

Unsetting these environment variables in the `op-scim-configmap` ConfigMap and redeploying is a workaround:
https://github.com/1Password/scim-examples/blob/a98b3d5626d03755c80c314f85d822274b8faea1/kubernetes/op-scim-config.yaml#L19-L20

ag-adampike:
This PR fixes #229, and projects the certificate and key files into the container (this was previously missing from the Deployment spec).

It also uses a [TLS Secret type](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) for the Secret where the certificate and key files are stored.

Finally, it amends the instructions to re-enable a custom cert by setting the `OP_TLS_CERT_FILE` and `OP_TLS_KEY_FILE` environment variables using `kubectl`.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -282,11 +282,24 @@ In this configuration, 1Password SCIM bridge will listen for unencrypted traffic
 
 #### Manually-Provided Key/Certificate
 
-Alternatively, you can create new secrets containing your key and certificate files, which can then be used by the SCIM bridge. This will also disable Let's Encrypt functionality.
+Alternatively, you can create a TLS Secret containing your key and certificate files, which can then be used by your SCIM bridge. This will also disable Let's Encrypt functionality.
+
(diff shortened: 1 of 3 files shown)


alicethorne-ab (approved):
Good catch! Approving.

black-bryan (approved):
Amazing Pike, good catch!

PR #232, opened 2023-03-08
Fix typo in tfvars

Rickerd0613:


scottisloud (approved):
Good catch @Rickerd0613 

no feedback ;) LGTM!

PR #233, opened 2023-04-21
Bump to v2.8.0, include new configuration options

alicethorne-ab:
Bumps to v2.8.0, and adds several new configuration options (with documentation):

* `OP_CONFIRMATION_INTERVAL` - allows users to set the interval of the automatic provisioning confirmation watcher
* `OP_DNS_CHALLENGE_CONFIG_FILE` - allows users to set the path for Let's Encrypt `DNS-01` challenges, with an example file included in `dns01-example.json`

NOTE: The `DNS-01` functionality is only being "soft launched" here. Further documentation on how to set this up will be refined at a later date.


dns01-example.json
grellyd, line 1: How do you feel about dns01.example.json?

grellyd (approved):
Thanks Alice!

chasdl (dismissed):
Looks good! Should we consider adding any documentation for the `OP_CONFIRMATION_INTERVAL`?

PR #248, opened 2023-08-14
246 docker swarm self managed tls reverts to letsencrypt Fix

black-bryan:
Better handling of certificate files for self-managed TLS deployment option

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -125,7 +125,7 @@ docker stack deploy -c docker-compose.yml -c gw-docker-compose.yml op-scim
 ```
 Learn more about [connecting Google Workspace to 1Password SCIM Bridge](https://support.1password.com/scim-google-workspace/).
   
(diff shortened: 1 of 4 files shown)


scottisloud:
I followed the revised deployment in a single-node swarm and still ended up with a Let's Encrypt cert. I'll give this another shot tomorrow with a freshly-provisioned swarm, since I reused a node that I've thrashed around somewhat, so it could easily be user error.

scottisloud:
After following the slightly-revised docs, my issue was user error, which I think the recent commits should help mitigate. Can confirm that, as written (and when the user is being attentive!), this results in the desired bridge with user-provided cert!

ag-adampike (approved):
Awesomesauce. An overlay is definitely much cleaner change for this use case and also very useful for the new Docker deployment example we're working on. :-)

LGTM! 🎉

PR #252, opened 2023-08-28
Docker beta deployment example improvements

ag-adampike:
This PR improves on the beta Docker Swarm deployment example (`/beta/docker`)

Rejigged setup workflow:
- move `OP_TLS_DOMAIN` declaration to `scim.env`
- remove redundant `scim.env` entries
- remove redundant example references
- added configs for advanced options

README improvements:
- typos, grammar
- clarify some steps
- add hints to helpful 1Password tooling
- add more links to Docker docs
- better explanations of some Docker concepts
- `--raw` flag in logs command for more more legible log output
- added advanced setup topics in appendix
- moar heading emoji 🎉😍

### DIFF
--- a/beta/docker/README.md
+++ b/beta/docker/README.md
@@ -5,7 +5,7 @@ This example describes how to deploy 1Password SCIM Bridge as a stack using [Doc
 ## In this folder
 
 - [`README.md`](./README.md): the document that you are reading. 👋😃
-- [`compose.template.yaml`](./compose.template.yaml): a [Compose specification](https://docs.docker.com/compose/compose-file/) format [Compose file](https://docs.docker.com/compose/compose-file/03-compose-file/) for 1Password SCIM Bridge.
+- [`compose.template.yaml`](./compose.template.yaml): a [Compose Specification](https://docs.docker.com/compose/compose-file/) format [Compose file](https://docs.docker.com/compose/compose-file/03-compose-file/) for 1Password SCIM Bridge.
 - [`compose.gw.yaml`](./compose.gw.yaml): an [override configuration](https://docs.docker.com/compose/multiple-compose-files/merge/) to merge the configuration necessary for customers integrating with Google Workspace.
 - [`redis.conf`](./redis.conf): a [Redis configuration file](https://redis.io/docs/management/config/) to load the Redis cache with the base configuration required for SCIM bridge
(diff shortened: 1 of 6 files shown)


beta/docker/README.md
```diff
@@ -48,41 +53,78 @@ All following steps should be run on the same computer where you are already usi
...
 
    ```sh
    docker swarm init # --advertise-addr 192.0.2.1
    ```
 
    > **Note**
    >
-   > If Docker returns an error about multiple network interfaces, uncomment the `--advertise-addr` parameter (delete `#`) and replace the example IP (`192.0.2.1`) with the IP address of a network interface on which the server can be reached externally.
+   > If Docker returns an error about multiple network interfaces, uncomment the `--advertise-addr` parameter (delete
+   > `#`) and replace the example IP (`192.0.2.1`) with the IP address of a network interface on which the server can
```
scottisloud, line 109: Typically when I've done this the default network interface will have both a private and public IP. I always sit and scratch my dumb head for a moment trying to think about which one to use here. Can we state explicitly which IP should be used or provide some guidance ("choose the external IP address") or whatever makes sense?

ag-adampike: I punted to the Docker Swarm docs for this in the very next lines, but perhaps I can clarify the explanation here a bit and expand the link text.

ag-adampike: How is [this](https://github.com/1Password/scim-examples/pull/252/commits/a75f3ef93873f3763b5f26349a525a9a3fa8e955)?

scottisloud: I don't know if the Docker docs nor your changes necessarily help me pick one of the two IP addresses assigned to the primary network interface (in my experience there are always two), but I might be an edge-case on account of being very prone to overthinking so this is probably fine.

ag-adampike: Thanks. Maybe I'll take another whack at it just the same. 👍

ag-adampike: I've expanded it significantly in [this commit](https://github.com/1Password/scim-examples/pull/252/commits/1a775b60b8e80064df16569dfee12856d69c5dae).

It's moot for the generic deployment example anyway, since we're recommending a single node. It just needs to be an IP address that another node can use to reach this one.

scottisloud (approved):
All concerns addressed. Just attempted a non-GW deployment with Let's Encrypt and it was a breeze. SHIP IT

PR #256, opened 2023-09-15
Kubernetes deployment example improvements

ag-adampike:
- introduces `runAs` and `fsGroup` spec to run Pods in context of respective container users for SCIM bridge and Redis
- run Pods as non-root user for best security practices
- mount Secrets in default location to reduce required config (removed need to set several environment variable values)
- refactor `REDIS_ARGS` into native `redis.conf` file (saves a silly warning line on boot, mounts the config file into the container, subjectively more pleasing than a list of args jammed into an environment variable value 😜)
- turn off snapshots by default in Redis config (unless the snapshots are being save by another process, these are redundant, lost at the end of a Pod's lifecycle, and never used)
- mount Redis data volume in Pod (if a manual snapshot is created, or automatic backups _are_ desired for some reason, customers can fetch the snapshot from this volume using another Pod container)
- update ConfigMap for SCIM bridge (corrections and clarifications)
- change relevant README.md entries, add instructions for cred rotation

In my testing, this applies with no fuss against an existing deployment. Tested on DigitalOcean and Azure.

To test:
- deploy 1Password SCIM bridge from the `master` branch
- checkout this branch
- apply the new manifests from this branch
- test functionality


black-bryan (dismissed):
LGTM!

black-bryan (approved):
LGTM!

PR #260, opened 2023-09-25
Fixed curl endpoint in step 5

erik-pineiro-github:
The curl example to test the SCIM bridge users endpoint was missing the `/scim/` portion which would result in the command not working.


erik-pineiro-github:
@ag-adampike mentioned that `/scim` endpoints are aliased and can be used or left out with the same result. After testing, I can confirm. I will leave this PR in here in case this sparks any ideas of changes we'd want to make around the surrounding copy.

grellyd:
As the existing command works, closing. Feel free to re-open if you wish to fix the wording to denote the optionality.

PR #262, opened 2023-10-13
Update Kub-recommended resources documentation

Hemal1P:
Updating documentation for "Very High Volume Deployment" category.

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -269,11 +269,11 @@ The best practices for Kubernetes resource limits are:
 
 This works well in the deployment scenario where the 1Password SCIM bridge is the only pod in the cluster, so that it has a priority of service and can consume all the resources in the cluster without affecting the performance of other pods.
 
-If you are deploying the SCIM bridge to a cluster with other production services, we recommend using the following resources depending on the size of your deployment.
+If you are deploying the SCIM bridge to a cluster **without** other production services, we recommend setting the following CPU and memory resources:
 
 > **Note**
 >
-> We are recommending not to set CPU limits when you expect performance. However you can set CPU limits if you have a good usecase to manage resources. It can be modified in your deployment based on our generic example and its expected to work in most circumstances.
+> We recommend to not set CPU limits when you expect higher performance. However, you can set CPU limits to manage resources if needed. This can be modified in your deployment based on our generic example, and its expected to work in most circumstances.
 
 ```yaml
 requests:
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -216,7 +216,9 @@ If you regenenerate credentials for your SCIM bridge:
 
 ## Appendix: Resource recommendations
 
-The default resource recommendations for the SCIM bridge and Redis deployments are acceptable in most scenarios, but they may fall short in high-volume deployments where a large number of users and/or groups are being managed. We strongly recommend increasing the resources for both the SCIM bridge and Redis deployments.
+The default resource recommendations for the SCIM bridge and Redis deployments are acceptable in most scenarios, but they may fall short in high-volume deployments where a large number of users and/or groups are being managed. We strongly recommend increasing the resources for both the SCIM bridge deployment.
+
+Once all users and groups have been provisioned, it is safe to lower these resources back to their defaults for everyday use of the SCIM bridge.
 
 | Expected Provisioned Users | Resources                   |
 | -------------------------- | --------------------------- |
@@ -252,7 +254,6 @@ requests:
   memory: 512M
 
 limits:
-  cpu: 1000m
   memory: 1024M
 ```
 
@@ -269,6 +270,10 @@ The best practices for Kubernetes resource limits are:
 This works well in the deployment scenario where the 1Password SCIM bridge is the only pod in the cluster, so that it has a priority of service and can consume all the resources in the cluster without affecting the performance of other pods.
 
 If you are deploying the SCIM bridge to a cluster with other production services, we recommend using the following resources depending on the size of your deployment.
+
+> **Note**
+>
+> We are recommending not to set CPU limits when you expect performance. However you can set CPU limits if you have a good usecase to manage resources. It can be modified in your deployment based on our generic example and its expected to work in most circumstances.
 
 ```yaml
 requests:
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -1,6 +1,6 @@
 # Deploy 1Password SCIM Bridge on Kubernetes
 
-*Learn how to deploy 1Password SCIM Bridge on a Kubernetes cluster.*
+_Learn how to deploy 1Password SCIM Bridge on a Kubernetes cluster._
 
 > **Note**
 >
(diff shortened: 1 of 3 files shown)


kubernetes/README.md
```diff
@@ -1,6 +1,6 @@
 # Deploy 1Password SCIM Bridge on Kubernetes
 
-*Learn how to deploy 1Password SCIM Bridge on a Kubernetes cluster.*
+_Learn how to deploy 1Password SCIM Bridge on a Kubernetes cluster._
```
chasdl, line 3: These autoformatting changes are actually changing bolded statements to italicized, which we don't want. Can we change back to using `*` ?

kubernetes/README.md
```diff
@@ -103,7 +103,7 @@ The [`op-scim-bridge` Service](./op-scim-service.yaml) creates a public load bal
 
 After the DNS record has propagated, set the `OP_TLS_DOMAIN` environment variable to enable the CertificateManager component.
 
-*Example command:*
+_Example command:_
```
chasdl, line 106: Same comment as above

kubernetes/README.md
```diff
@@ -117,7 +117,7 @@ You can test the connection to your SCIM bridge and view status and logs in a we
 
 You can also test your SCIM bridge by sending an authenticated SCIM API request.
 
-*Example command:*
+_Example command:_
```
chasdl, line 120: Same comment as above

kubernetes/README.md
```diff
@@ -349,7 +358,7 @@ If you prefer to use an external Redis cache, omit the the `redis-*.yaml` files
 
 You can set `OP_REDIS_URL` directly to reboot SCIM bridge and connect to the specified Redis server:
 
-*Example command:*
+_Example command:_
```
chasdl, line 361: Same comment as above

kubernetes/README.md
```diff
@@ -210,33 +210,34 @@ If you regenenerate credentials for your SCIM bridge:
    ```
 
    Kubernetes automatically updates the credentials file mounted in the Pod with the new Secret value.
+
 3. Test your SCIM bridge using the new bearer token associated with the regenerated `scimsession` file.
 4. Update your identity provider configuration with the new bearer token.
 
 ## Appendix: Resource recommendations
 
 The default resource recommendations for the SCIM bridge and Redis deployments are acceptable in most scenarios, but they may fall short in high-volume deployments where a large number of users and/or groups are being managed. We strongly recommend increasing the resources for both the SCIM bridge and Redis deployments.
```
chasdl, line 219: Since the performance improvements we've made (see https://docs.google.com/document/d/1cD5GY8kNm-RwqqA1sklwzx77yoknmGV-bfaTEvUv9sg/edit), we no longer need to recommend customers bump resources for their Redis deployments. So this statement can read:

> We strongly recommend increasing the resources for the SCIM bridge deployment.

I would also like to add a comment about deploy resources vs. normal everyday use of the SCIM bridge after this comment. Something like:

> Once all users and groups have been provisioned, it is safe to lower these resources back to their defaults for everyday use of the SCIM bridge.

kubernetes/README.md
```diff
@@ -245,30 +246,38 @@ Note that these are the recommended `requests` and `limits` values for both the
...
+
+If you are deploying the SCIM bridge to a cluster with other production services, we recommend using the following resources depending on the size of your deployment.
+
+```yaml
+requests:
+  cpu: 1000m
+  memory: 1024M
+
+limits:
+  memory: 1024M
```
chasdl, line 279: Is there a reason we left off the CPU limit of `cpu: 2000m`?

chasdl (changes requested):
Looking pretty good! A few comments.

kubernetes/README.md
Hemal1P: Comparing with our [main branch](https://github.com/1Password/scim-examples/tree/master/kubernetes) the text initially represented with *xxx* as italics are now adapted as  _xxx_ by editor. Do we want to make it bold there?

kubernetes/README.md
Hemal1P: Thanks. Thats good callout. <br/>
For the second suggestion, I thought we have lines  after the **Default** section but its right to add on top section as well in case one skips to read further.

kubernetes/README.md
Hemal1P: So as part of no CPU limit strategy we remove it from limits to in order to decrease occurrences of CPU throttling and make full use of available CPU.

kubernetes/README.md
```diff
@@ -245,30 +248,41 @@ Note that these are the recommended `requests` and `limits` values for both the
...
 
-  limits:
-    cpu: 2000m
-    memory: 2048M
-  ```
+If you are deploying the SCIM bridge to a cluster with other production services, we recommend using the following resources depending on the size of your deployment.
+
+> **Note**
+>
+> We are recommending not to set CPU limits when you expect performance. However you can set CPU limits if you have a good usecase to manage resources. It can be modified in your deployment based on our generic example and its expected to work in most circumstances.
```
alicethorne-ab, line 276:
```suggestion
> We recommend to not set CPU limits when you expect higher performance. However, you can set CPU limits to manage resources if needed. This can be modified in your deployment based on our generic example, and its expected to work in most circumstances.
```

alicethorne-ab (commented):
Minor nitpick, but otherwise looking good.

chasdl (approved):
Looking good. Please apply Alice's suggestion for the updated copy, but otherwise I'm happy with this. Please check with @ag-adampike before merging as well.

ag-adampike:
Even with other workloads in a cluster, it is still recommended (in most cases) to remove CPU limits in Pod specs so that available CPU cycles are not potentially wasted, and to set both memory requests and limits and set them equivalently for predictable memory management. This goes the same for other resources in a cluster in general.

References:

- [Remove your CPU Limits | by Shon Lev-Ran | Directeam](https://medium.com/directeam/kubernetes-resources-under-the-hood-part-3-6ee7d6015965)
- [What Everyone Should Know About Kubernetes Memory Limits, OOMKilled Pods, and Pizza Parties | Robusta](https://home.robusta.dev/blog/kubernetes-memory-limit)

If a customer requires some bespoke configuration for a production cluster that is running other workloads, this base example can still be modified to suit their needs on their end.

I've created a commit (https://github.com/1Password/scim-examples/commit/e542ca93095f27b7dd8ebfcff825319f7ac438ca) in a branch off this one that modifies our base deployment example Deployment manifests this way:

```yaml
# ...
          resources:
            requests:
              cpu: 125m
              memory: 512M
            limits:
              memory: 512M
# ...
```

This removes a line each from two manifest files (even easier to add your own spec 😉), meets best practices for speccing resources for Kubernetes by default, and simplifies instructions to vertically scale the SCIM bridge Deployment spec as needed.

ag-adampike (approved):
LGTM! Thanks you for accomodating my requests! ;-)

kubernetes/README.md
Hemal1P: That reads better. Thanks. 😄  
Changes applied.

ag-adampike (approved):

PR #264, opened 2023-10-17
Update AWS CloudFormation example code & yaml file

muhwagwa:
2 updates for the AWS CloudFormation SCIM bridge example

1. fix minor error on example code
2. add default value for WorkspaceActor & WorkspaceCredentials (it previously gave an error when these 2 parameters were not put in)

PR #265, opened 2023-10-18
Update AWS CloudFormation example code & yaml file

muhwagwa:
2 updates for the AWS CloudFormation SCIM bridge example

- fix minor error on example code
- add default value for WorkspaceActor & WorkspaceCredentials (it previously gave an error when these 2 parameters were not put in)

Contains same changes from #264 due to non-signed commits in #264.

### DIFF
--- a/beta/aws-ecsfargate-cfn/README.md
+++ b/beta/aws-ecsfargate-cfn/README.md
@@ -95,7 +95,7 @@ You can alternatively deploy 1Password SCIM Bridge with a single AWS CLI command
 *Example command:*
 
 ```sh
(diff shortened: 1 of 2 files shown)


ag-adampike (approved):
Excellent find. Thanks for taking this on! 🎉

Can confirm that the changes allow correct functionality of the example command and for validation to work as expected to enforce that either both Google Workspace parameter values are set (when using Google Workspace as the IdP) or neither are (if using any other IdP).

Deployment now works as expected using AWS CLI.

scottisloud (approved):
Good catch! Let's rock and roll! Approving

PR #266, opened 2023-10-31
Revise resource specifications and recommendations

ag-adampike:
Our engineering team responsible for developing our SCIM bridge container image has spent the last several months investigating and refining how SCIM bridge performs at scale. We now have updated guidelines for recommended resources that should be allocated to the SCIM bridge container when provisioning large numbers of users and groups to 1Password.

This PR revises our deployment examples to:

- ensure that the default resource allocations for the SCIM bridge and Redis containers on each respective platform align with our base recommendations
- include guidelines for vertically scaling at higher volumes of provisioning activity
- remove any references to scaling the Redis container and recommend always using the default resource specification

### DIFF
--- a/kubernetes/README.md
+++ b/kubernetes/README.md
@@ -216,96 +216,60 @@ If you regenenerate credentials for your SCIM bridge:
 
 ## Appendix: Resource recommendations
 
-The default resource recommendations for the SCIM bridge and Redis deployments are acceptable in most scenarios, but they may fall short in high-volume deployments where a large number of users and/or groups are being managed. We strongly recommend increasing the resources for both the SCIM bridge deployment.
+The 1Password SCIM Bridge Pod should be vertically scaled when provisioning a large number of users or groups. Our default resource specifications and recommended configurations for provisioning at scale are listed in the below table:
 
-Once all users and groups have been provisioned, it is safe to lower these resources back to their defaults for everyday use of the SCIM bridge.
+| Volume    | Number of users | CPU   | memory |
+| --------- | --------------- | ----- | ------ |
+| Default   | <1,000          | 125m  | 512M   |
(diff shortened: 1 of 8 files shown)


black-bryan (approved):
🎉 LGTM

kubernetes/README.md
```diff
@@ -427,3 +391,13 @@ kubectl set env deploy/op-scim-bridge OP_PING_SERVER=1
 ```
 
 No other endpoints (such as `/scim`) are exposed through this port.
+
+## Troubleshooting
+
+When updating cluster resources in place, Kubernetes will create a new revision alongside an existing workload before replacing the existing revision. If there are not enough available resources in the cluster on which to schedule both revisions, the update can hang until they become available. This can happening when vertically scaling a Deployment or updating to a new SCIM bridge image version.
+
+If there are enough resources in the cluster available to schedule the new Pod revision, you can restart the Deployment to evict the existing Pod, free up the resources, and allow the new Pod to start:
```
Penderworth, line 399:
```suggestion
If there are enough resources in the cluster available to schedule the new Pod revision, you can restart the deployment to evict the existing Pod, free up the resources, and allow the new Pod to start:
```

Penderworth (changes requested):
Just two small updates and this is ready to go from the wording side of things.

kubernetes/README.md
ag-adampike: This one is intended. Kubernetes resource types are proper nouns: Deployment, Pod, StatefulSet, etc. I am referring to the Deployment object here specifically.

Penderworth: Aha, TIL!

Penderworth (approved):

PR #275, opened 2023-12-19
Fix template download commands

ag-adampike:
This PR fixes the incorrect links in the template file download commands for the Azure Container Apps deployment example, and refactors the commands.


muhwagwa (approved):
LGTM!

PR #289, opened 2024-01-30
Add tip callout to main README

Penderworth:
I'll be updating some of the other READMEs if we like the style. My Markdown editor just added support for rendering them so I noticed there are more types than I thought. GitHub calls these "alerts" and their documentation is here: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

What it looks like in preivew:

Screenshot 2024-01-30 at 9 20 07 AM


Penderworth:
@black-bryan This is a small change if you have time to review it. I wanted to get readers' attention a bit more here because _some_ may not have visited our KB first and the main SCIM page has a lot of info that's helpful to know.

black-bryan (approved):
❤️ the different GH markdown `alerts`. Nice find, great way to highlight our SCIM KB first.

PR #294, opened 2024-03-08
Fixed formatting issues in Azure Container Apps README

Penderworth:
We missed a few formatting issues in the previous PR, so this fixes them. It also uses unordered lists for the nested ordered list items, since those render different on GitHub than I expected and the i,ii,iii,iv etc. rendering doesn't look very good in context.


black-bryan (approved):
Looks good! Thanks for quickly catching these.

PR #303, opened 2024-05-03
bump to 2.9.4

Hemal1P:
Bump to v2.9.4

Changelog [link](https://app-updates.agilebits.com/product_history/SCIM#v209031)


chasdl (approved):

ag-adampike:
The only slight reservation is that we will need to document the new lower limit for the Confirmation Watcher interval, but it may be just as well to document that in a separate issue/PR.

If you're happy with that, LGTM! :-)

ag-adampike (approved):

PR #316, opened 2024-07-12
Add App Runner SCIM bridge example

plttn:
This adds App Runner as an available deployment option in beta.

### Note

Due to the fact that ECR is required, this is being submitted as draft and cannot be moved to final review until such time that 1Password SCIM Bridge is published on public ECR.

### DIFF
--- a/beta/aws-apprunner-cfn/README.md
+++ b/beta/aws-apprunner-cfn/README.md
@@ -1,0 +1,242 @@
+# Deploy 1Password SCIM Bridge with AWS CloudFormation on App Runner
+
+This example describes how to deploy 1Password SCIM Bridge using [AWS CloudFormation](https://docs.aws.amazon.com/cloudformation/index.html) on Amazon App Runner
+
+## In this folder
+
+- [`README.md`](./README.md): the document that you are reading. 👋😃
+- [`op-scim-bridge.yaml`](./op-scim-bridge.yaml): a [CloudFormation template](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-whatis-concepts.html#cfn-concepts-templates) for 1Password SCIM Bridge.
+
+## Overview
+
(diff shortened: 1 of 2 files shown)


plttn:
Making ready for review per internal discussions with requirement to upload to ECR per customer.

ag-adampike:
Added some sweeping changes to the template in bf28001. Most changes do not change any functionality. One thing I forgot to include in the commit message was that I reordered the resources in a logical order for readability and clarity.

We'll still need to account for pulling a specific image from ECR and may want to introduce a parameter to reference it in the AppRunnerService resource, and update the docs to include some steps about pushing our public image to a private registry, or consider if and how we might automate this as part of the template.

plttn:
> Added some sweeping changes to the template in [bf28001](https://github.com/1Password/scim-examples/commit/bf28001a2a7e732dfa4ee18d7ec004ea1b988933). Most do not change any functionality. One thing I forgot to include in the commit message was that I reordered the resources in a logical order for readability and clarity.
> 
> **We'll still need to account for pulling a specific image from ECR and may want to introduce a parameter to reference it in the AppRunnerService resource, and update the docs to include some steps about pushing our public image to a private registry, or consider if and how we might automate this as part of the template.**

I'm not thrilled with what the flow would look like (there's additional stuff that would have to be spun up in AWS [CodeBuild]). I think documenting how to pull/push the docker image to an ECR private is going to best.

ag-adampike:
> I'm not thrilled with what the flow would look like (there's additional stuff that would have to be spun up in AWS [CodeBuild]). I think documenting how to pull/push the docker image to an ECR private is going to best.

I've already got it worked out in a draft version of the template that also sets up a private ECR repo and a CodeBuild project that pulls from Docker Hub and pushes to the private ECR repo. It completes all these steps in a _much_ shorter time than it takes just to deploy ElastiCache.

Since an ECR image is required for App Runner, I'd prefer to include this functionality with the template, rather than leave it up to a customer to have to do before deployment.

I'll share that in the next commit for consideration. :-)

accraw:
App Gateway does a better job.

PR #319, opened 2024-07-18
Update aca-op-scim-bridge.yaml

Owemeone:
When using a different naming convention for the Container App, the cli start complaining about mismatches in name. Dropping the name from the yaml config, keeps the name as used during creation.

I don't really care about the container names within the App, so left those as-is.


accraw:
Thank you for taking the time to open this PR. We have made the decision to deprecate this deployment to reduce confusion, please use the ARM deployment instead: https://github.com/1Password/scim-examples/tree/main/azure-container-apps-arm

PR #314, opened 2024-07-05
change resource recommendation for memory-very high

muhwagwa:
Edit resource recommendation section of Azure Container Apps deployment.
For `high volume` > `memory`, it should be 2.0Gi per [this docs](https://arc.net/l/quote/zjmnltwb).

PR #321, opened 2024-07-18
Small resource spec fix for ACA very high volume deployment within the Azure portal steps

black-bryan:
Small fix to align the very high volume resource specs steps for making changes in the Azure GUI portal to match the az commands shown in the Advanced.md. 

These specs were recently updated in [PR314](https://github.com/1Password/scim-examples/pull/314), however the very high volume written steps within the Azure Portal were missed in the previous PR. 

Memory should be spec'd at 2.0Gi.


muhwagwa (approved):
LGTM

PR #325, opened 2024-07-30
Fix to command in GCP Cloud Run Google Workspace secret README

black-bryan:
There is a small issue where the existing README had an incorrect command with extra `&&` at the end of the command, likely a small copy/paste aspect that was missed earlier. This command is only used when Google Workspace is the IdP, on a GCP Cloud Run SCIM bridge. 

Tested the updated command with a customer today.


ag-rdoucette (approved):
Hit the same issue and you beat me to updating this!

LGTM!

PR #330, opened 2024-08-23
Fixed analytics name in Azure ARM

Penderworth:
This name was incorrect and we noticed it during the AKB docs.


muhwagwa (approved):
LGTM!

PR #338, opened 2024-09-13
Fix scaling in Azure Resource Manager template

### LINKED ISSUE #337: Problem: Azure Container App ARM Deployment Fails to Set Minimum Replicas to 1
We noticed an issue with our **Azure Container App ARM deployment** where it fails to set the **minimum replicas** to **1**. This causes the **1Password SCIM bridge** to potentially scale down to **0** replicas, leading to service disruptions for the **ConfirmationWatcher**, which then requires a bearer token to start back up.

## Impact
- The **SCIM bridge** scales down to **0 replicas**, causing it to stop.
- **Manual intervention** is required to pass a bearer token to start the ConfirmarionWatcher.
  
---

## Workaround: Setting the Minimum Replicas via Azure Portal UI

Follow the steps below to manually configure the **minimum replicas** to **1** using the **Azure Portal**:

### Step-by-Step Instructions

1. **Log in to the Azure Portal**:
   - Open the [Azure Portal](https://portal.azure.com) and sign in with your credentials.

2. **Navigate to Your Container App**:
   - In the left sidebar, click **"Container Apps"**.
   - Find and select your **Container App** running the SCIM bridge.

3. **Access the Scale Settings**:
   - In the **Container App Overview** page, go to the **"Scale"** section under **Application**.

4. **Modify the Scaling Rules**:
   - Locate the **"Replica count"** settings.
   - Set the **Minimum replicas** field to **1** to ensure at least one instance of the SCIM bridge is always running.

5. **Save the Changes**:
   - Once you've updated the minimum replicas, click **"Save"** to apply the changes.

6. **Verify the Configuration**:
   - Return to the **Overview** page for your container app.
   - Confirm that the **Replicas** count shows at least **1 replica** running.

---

### Example of Replica Count Setting in Azure Portal

Here’s what the **Scale** settings page should look like:

| Setting           | Value |
|-------------------|-------|
| Minimum Replicas   | 1     |
| Maximum Replicas   | 1     |

---

## Notes:
- This is a **temporary workaround** to prevent the SCIM bridge from scaling down to **0 replicas**.
- The change will take effect immediately once applied in the Azure Portal.

---

## Expected Behavior
- The SCIM bridge will always maintain **1 replica**, preventing it from scaling down to **0** and ensuring continuous operation.

## Actual Behavior
- Without the fix, the SCIM bridge scales down to **0 replicas**, causing the **ConfirmationWatcher** to stop, and requiring a manual restart with a bearer token.

### Environment
- **Azure Container Apps**: ARM deployment template

---

Please advise on a long-term solution to resolve this issue in the ARM deployment template.

ag-rdoucette:
Updated the scaling from 0 to 1

### DIFF
--- a/azure-container-apps-arm/aca-op-scim-bridge-template.json
+++ b/azure-container-apps-arm/aca-op-scim-bridge-template.json
@@ -174,7 +174,12 @@
                             "name": "credentials",
                             "storageType": "Secret"
                         }
-                    ]
+                    ],
+                    "scale": {
+                        "minReplicas": 1,
+                        "maxReplicas": 10,
+                        "rules": []
+                    }
                 }
             }
         },
--- a/azure-container-apps-arm/aca-op-scim-bridge-template.json
+++ b/azure-container-apps-arm/aca-op-scim-bridge-template.json
@@ -177,8 +177,7 @@
                     ],
                     "scale": {
                         "minReplicas": 1,
-                        "maxReplicas": 10,
-                        "rules": []
+                        "maxReplicas": 1
                     }
                 }
             }


ag-adampike (approved):
Tested the fix by applying the template using existing resource names. Works as expected! 🎉

We might consider including instructions to do the same for anyone who deployed using the current template.

We could also tackle that in a new PR, so I've approved it if you prefer to merge as is. :-)

PR #339, opened 2024-09-17
Bump to v2.9.6 in CFN & Docker examples

ag-adampike:
Version bump to v2.9.6 for our Docker and CloudFormation template deployment examples.


ag-rdoucette (approved):
Looks like a basic change on the version in 3 files.

LGTM.

PR #340, opened 2024-09-19
Specify ARM64 architecture in Terraform deployment example

ag-rdoucette:
1Password SCIM Bridge [version 2.9.0](https://releases.1password.com/provisioning/scim-bridge/#1password-scim-bridge-2.9.0) introduced ARM64 support (Redis already includes ARM64 support). Our Terraform deployment example does not specify a runtime platform and defaults to x86_64 architecture. ARM64 architecture is preferred for lower cost with no expected performance degradation.

This PR:

- adds a `runtime_platform` block in the ECS task definition
- specifies `cpu_architecture` as `ARM64` and `operating_system_family` as `LINUX`

In testing, the ECS task starts up as expected with the correct `runtime_platform` specification.


ag-adampike (approved):
LGTM!

PR #350, opened 2025-02-26
Remove AKS reference

plttn:
Now that `@scim-deploy-azure` points to ACA, the callout here doesn't make sense.


black-bryan:
Let's actually update the link to here: https://support.1password.com/cs/scim-deploy-azure-kubernetes/

accraw (approved):

PR #352, opened 2025-03-14
Update AWS ALB SSL policy to `ELBSecurityPolicy-TLS13-1-2-Res-2021-06`

reedloden:
Change the SSL policy used for AWS ALBs from `ELBSecurityPolicy-TLS13-1-2-2021-06` to `ELBSecurityPolicy-TLS13-1-2-Res-2021-06`.

This removes the following (insecure) cipher suites:
- ECDHE-ECDSA-AES128-SHA256
- ECDHE-RSA-AES128-SHA256
- ECDHE-ECDSA-AES256-SHA384
- ECDHE-RSA-AES256-SHA384

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/describe-ssl-policies.html#tls-security-policies

### DIFF
--- a/aws-ecsfargate-terraform/main.tf
+++ b/aws-ecsfargate-terraform/main.tf
@@ -270,7 +270,7 @@ resource "aws_lb_listener" "https" {
   load_balancer_arn = aws_alb.op_scim_bridge.arn
   port              = 443
   protocol          = "HTTPS"
-  ssl_policy        = "ELBSecurityPolicy-TLS13-1-2-2021-06"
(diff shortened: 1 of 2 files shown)


plttn (approved):
Based on my research, I don't believe this will cause any negative impacts for IdPs. Approving, and will make an internal note for myself and colleagues if we experience any issues with customers. 

Thanks again for flagging this!

plttn:
With that said, I'll have to make a new PR and make you a co-author, as we require all commits to be signed here.

PR #361, opened 2025-05-19
fix: beta/google-cloud-run use a service account with minimal permiss…

RichardoC:
…ions

Google recommend not using this service account as it's has more permissions than are typically required [[1]](https://cloud.google.com/compute/docs/access/service-accounts#default_service_account)

This PR updates the cloud run setup to use a dedicated service account, and then only grant the required permissions to it instead.


ag-adampike:
Hey @RichardoC! Thanks for this (pretty freakin' sweet) PR, and for taking the time to include accompanying documentation changes and the linked reference. 😍

This is indeed good guidance corroborated as a best practice in other Google Cloud documentation[^1][^2] principle of least privilege.

I have to review the changes and add some for customers that use SCIM bridge with Google Workspace as their identity provider. That case adds some potential complexity: because a SCIM client application is not available for Workspace, our bespoke provisioning integration _always_ requires a Google Cloud service account key for this purpose. 🙃

Not getting that bit tangled up was one honest reason for using the default Compute Engine service account instead of a dedicated service account in the first place.

In any case, I'll be able to have a more thorough look at this next week and hopefully merge it shortly after.

Thanks again! 💙

[^1]: [Best practices for using service accounts](https://cloud.google.com/iam/docs/best-practices-service-accounts#single-purpose)
[^2]: [Introduction to service identity](https://cloud.google.com/run/docs/securing/service-identity#default_service_account)

RichardoC:
Hi @ ag-adampike, any updates?

PR #362, opened 2025-05-28
Added Redis Dependency for SCIM and Redis Healthcheck in Cloud Run

ag-rdoucette:
# Pull Request Description

This PR updates the `op-scim-bridge.yaml` and `op-scim-bridge-gw.yaml` configurations for the **Cloud Run 1Password SCIM bridge** to address timing issues where the `scim` container attempts to connect to `redis` before it is fully operational.

## Changes

- Added `run.googleapis.com/container-dependencies` annotation to both `op-scim-bridge.yaml` and `op-scim-bridge-gw.yaml` to ensure the `scim` container starts only after the `redis` container is ready.
- Added a `startupProbe` (TCP check on port `6379`, `5s` initial delay, `10s` period, `3` failure threshold, `1s` timeout) to the `redis` container in both YAMLs to confirm Redis readiness.

### DIFF
--- a/beta/google-cloud-run/google-workspace/op-scim-bridge-gw.yaml
+++ b/beta/google-cloud-run/google-workspace/op-scim-bridge-gw.yaml
@@ -10,6 +10,7 @@ spec:
         autoscaling.knative.dev/maxScale: "1"
         run.googleapis.com/cpu-throttling: "false"
         run.googleapis.com/startup-cpu-boost: "true"
+        run.googleapis.com/container-dependencies: '{"scim":["redis"]}'
     spec:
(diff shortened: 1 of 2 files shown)


ag-adampike (approved):
Tested and working as expected!

A more robust "`exec`-style" health check for Redis (similar to our other deployment examples) would be preferred to enable Cloud Run to establish the dependency order, but sadly isn't available for Cloud Run. In the absence of that, the TCP readiness check parameters should be sufficient to ensure that Redis is actually healthy and appeared to consistently do the trick in our testing.

LGTM! 🎉

PR #382, opened 2025-10-02
Fix broken Redis image for DigitalOcean App Platform example

ag-adampike:
Our deployment example for DigitalOcean App Platform uses the Bitnami Redis container image in its app spec. This image and its repository are no longer available on Docker Hub as part of planned changes to Bitnami's catalog (see https://github.com/bitnami/containers/issues/83267).

1Password SCIM Bridge does not require any specific Redis distribution, and all other deployment examples do not specify a repository and use the Docker Official "library" Redis image (https://hub.docker.com/_/redis).

In this PR, we

- switch to the Docker Official Redis image on Docker Hub
- move Redis configuration from environment variables to container run command to align with other deployment examples and produce slightly cleaner logs
- remove redundant env var config specific to Bitnami
- include the missing `save ""` configuration directive to disable redundant paging to disk
- carry out same changes in Google Workspace app spec file

### DIFF
--- a/do-app-platform-op-cli/google-workspace/op-scim-bridge-gw.yaml
+++ b/do-app-platform-op-cli/google-workspace/op-scim-bridge-gw.yaml
@@ -1,22 +1,15 @@
 name: op-scim-bridge
 services:
-  - envs:
-      - key: ALLOW_EMPTY_PASSWORD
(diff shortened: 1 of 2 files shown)


accraw (approved):
Tested it, works for me!

PR #386, opened 2025-10-28
Fix wildcard cert

mdmallardi:
Our Lets Encrypt certs have changed their domain order list to list the bare domain first.  This commit corrects the wilcard_cert call in this module to compensate for this change.  That said: This was the exact change for which we created this internal fork of this module.  We should reevaluate the need for this fork.


mdmallardi:
Oh gosh.  I just realized I opened this against the wrong repo. My apologies

PR #402, opened 2026-09-22
Support trusted proxies in all deployment examples

plttn:
## What this does

1Password SCIM Bridge rate limits requests per client IP address. Most of these examples put a TLS-terminating proxy in front of the bridge, so every request arrives from one address and the whole tenant gets rate limited as a single caller.

This sets `OP_SCIM_RATE_LIMIT_TRUSTED_PROXIES` across the examples so the bridge reads the real client address from `X-Forwarded-For`. Where the template already knows the proxy range, the value is set for you; where it can't be known, the variable ships commented out with an explanation.

## Per example

| Example | Value | How |
|---|---|---|
| `aws-ecsfargate-terraform` | Public subnet CIDRs | Computed — new `data "aws_subnet" "public"` lookup, passed through `templatefile()` |
| `aws-ecs-apigateway-cfn` | VPC CIDR, or new `TrustedProxies` parameter | `!If` on a new `HasTrustedProxies` condition |
| `azure-container-apps` (+ `-arm`) | Private + link-local ranges | New `env` on the bridge container |
| `google-cloud-run` | Private + link-local ranges | New `env` on the `scim` container |
| `do-app-platform-op-cli` | Private + link-local ranges | Added to the existing `envs` list |
| `docker`, `kubernetes`, `beta/kustomize` | Unset | Commented stub — these terminate TLS at the bridge, so it only applies behind an operator-added proxy |

The broad private + link-local default is only used on the managed platforms, where the platform ingress is the sole route to the container. It is deliberately **not** the default on Docker or Kubernetes, where the bridge can be reachable directly and a trusted range would let a caller spoof the header.

Google Workspace variant manifests were updated alongside their base manifests.

## Also included

`kubernetes/README.md` now explains how the Service's `externalTrafficPolicy` affects the source address the bridge sees. The default `Cluster` policy can replace the client address with a node address, which trusted proxies cannot recover. The manifest stays on the default — with the single replica these examples ship, `Local` would make the load balancer route only to the node running the Pod, and that tradeoff belongs to the operator.

## Two notes on the original plan

- `aws-ecs-apigateway-cfn` is API Gateway **v2 HTTP API** → VPC Link → ECS via Cloud Map. There is no NLB and no REST API, so the concern about an L4 hop stripping headers doesn't apply here.
- The ARM template's Google Workspace path inherits containers via `reference(...).template.containers`, so it picks up the new `env` automatically and needed no second edit.

## Verification

- `terraform validate` passes; all four `scim.json` placeholders are supplied.
- `cfn-lint` reports no new findings (the one `E1041` is pre-existing on `main`).
- `kustomize build` succeeds; `docker stack config` succeeds.
- All edited YAML and JSON parse.
- Both `ADVANCED.md` files were byte-identical before and remain so.

Not covered: `kubectl apply --dry-run` needs a live cluster, and no example was deployed. An end-to-end check — confirming two distinct source addresses are rate limited independently — is still worth doing on one deployment before release.

## Out of scope

The Redis contention behind the 503s (PROV-2441), the `OP_SCIM_RATE_LIMIT=0` messaging, root README warnings, and `deprecated/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


plttn:
not moving forward with this.

